PR #5, opened 2016-04-21
Opacity for the controls' background and fade-out

jeeb:
* Makes the background of the controls be 75% opaque (25% transparent).
* Adds IDs to some of the layout groups
* Change the time for the controls to be hidden from 1 second to 2.
* Rework the initialization, showing and hiding of the Android UI on-screen controls.

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.java
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.java
@@ -26,7 +26,7 @@ import java.io.OutputStream;
 public class MPVActivity extends Activity implements EventObserver {
     private static final String TAG = "mpv";
     // how long should controls be displayed on screen
-    private static final int CONTROLS_DISPLAY_TIMEOUT = 1000;
+    private static final int CONTROLS_DISPLAY_TIMEOUT = 2000;
 
     MPVView player;
     View controls;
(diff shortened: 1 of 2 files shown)


app/src/main/java/is/xyz/mpv/MPVActivity.java
```diff
@@ -268,22 +290,26 @@ public void updateDecoderButton() {
...
 
-    @Override public void run() {
+    @Override
+    public void run() {
         // use GONE here instead of INVISIBLE (which makes more sense) because of Android bug with surface views
         // see http://stackoverflow.com/a/12655713/2606891
-        controls.setVisibility(View.GONE);
-
-        int flags = View.SYSTEM_UI_FLAG_HIDE_NAVIGATION | View.SYSTEM_UI_FLAG_FULLSCREEN;
-        decorView.setSystemUiVisibility(flags);
```
xyzz: Why is this not called here anymore? this results in back/home/square buttons not going away when controls disappear

jeeb: OK, I misunderstood what these were doing as I haven't used a soft button Android device yet. Enabled them on my OPO and now I can see the issue.

jeeb:
Got an LGTM, merging outside the pull request system

Issue #11, opened 2016-04-22
stats.lua built-in alternative


xyzz:
stats.lua: https://github.com/Argon-/mpv-stats

When controls are shown and the option is enabled in developer settings ( #3 ) it should display some simple text overlay with all info written on it, easiest would be to implement it all in Java since we already have all the property retrieval functions.

xyzz:
WIP: https://github.com/mpv-android/mpv-android/tree/xyz/stats

this is waiting for settings menu #3

joshdekock:
How do you want to merge this? want me to make a patch, apply it to my settings branch, and then create another commit for the actual setting, or do you want to put the setting in with this branch (which would require #3 merged first)

jeeb:
Initial version merged.

PR #19, opened 2016-05-22
Add settings menu

joshdekock:
Add a settings menu, and fix the theme a little.
Settings that work:
- audio language
- subtitle language
- hardware decoding
- upscaling method
- downscaling method
- scale parameters (1 and 2)
- debanding
- custom vo
- custom ao
Settings that don't work:
- Show stats (there's no functionality yet)
- Show FPS (same as above)


joshdekock:
@jeeb has taken over the merging of this, so I'm going to close this PR.

joshdekock:
Can we just merge this? Firstly, It's blocking #11, and surely we can build on it over time rather than doing it all in one?

jeeb:
This looks good in general as you seem to have rebased+added file picker root. I will poke at it as soon as I have time and then for me it's good-to-go unless I find something glaring.

@xyzz , @sfan5 : if you have time please check it out and if there are no issues in your opinion => feel free to merge as well.

app/src/main/java/is/xyz/mpv/MPVView.kt
```diff
@@ -23,15 +24,71 @@ internal class MPVView(context: Context, attrs: AttributeSet) : GLSurfaceView(co
...
         val framesPerBuffer = am.getProperty(AudioManager.PROPERTY_OUTPUT_FRAMES_PER_BUFFER)
         val sampleRate = am.getProperty(AudioManager.PROPERTY_OUTPUT_SAMPLE_RATE)
         Log.v(TAG, "Device reports optimal frames per buffer $framesPerBuffer sample rate $sampleRate")
-        MPVLib.setOptionString("ao", "opensles:frames-per-buffer=$framesPerBuffer:sample-rate=$sampleRate")
+
+        var ao = "opensles:frames-per-buffer=$framesPerBuffer:sample-rate=$sampleRate"
+
+        // overrides
+
+        var opengl_custom = sharedPreferences.getString("opengl_custom", "")
```
sfan5: What's the purpose of allowing overriding of vo?
Anything except `opengl-cb` will simply not work.

app/src/main/java/is/xyz/mpv/SettingsActivity.kt
```diff
@@ -0,0 +1,224 @@
...
+    private fun setupActionBar() {
+        val actionBar = actionBar
+        actionBar?.setDisplayHomeAsUpEnabled(true)
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    override fun onIsMultiPane(): Boolean {
+        return isXLargeTablet(this)
```
sfan5: Is it <i>normal</i> (in Android Dev) to only have multi pane on x-large tablet devices?

If not, it would make sense to document why this choice was made.

joshdekock: Multi-pane wouldn't really fit on any other device, and the settings is based off the android studio template, so I guess I just assumed that it should be there, I don't actually have a tablet to test it out though.

app/src/main/java/is/xyz/mpv/MPVView.kt
xyzz: The purpose is allowing users to set custom options not present yet in the settings UI. Maybe it's better to make it append to vo instead of overriding, but then it would enable users to first set an option in the UI (e.g. video upscale) and then add the same option manually.

app/src/main/res/xml/pref_developer.xml
```diff
@@ -0,0 +1,15 @@
...
+        android:defaultValue="false"
+        android:key="show_stats"
+        android:summary="@string/pref_developer_show_stats_summary"
+        android:title="@string/pref_developer_show_stats_title" />
+
+    <CheckBoxPreference
+        android:defaultValue="false"
+        android:key="show_fps"
+        android:summary="@string/pref_developer_show_fps_summary"
+        android:title="@string/pref_developer_show_stats_title" />
```
sfan5: should be `@string/pref_developer_show_fps_title`

sfan5:
Tested and seems to work! :+1:

app/src/main/java/is/xyz/mpv/MPVView.kt
joshdekock: I think it's just better to allow fully overriding it for now, and then potentially remove the custom option in the future

app/src/main/res/xml/pref_developer.xml
joshdekock: Fixed.

PR #24, opened 2016-07-31
Playback Statistics Overlay

jeeb:
* Adds @xyzz 's original playback overlay.
* Plugs the playback overlay into @joshdekock 's settings entry available in the settings screen.


jeeb:
Test build available [here](https://github.com/mpv-android/mpv-android/files/392662/mpv-android-2016-07-31-with-stats.zip).

Rename zip to apk after downloading. Github is a dummie for not letting to upload files with the extension apk.

PR #28, opened 2016-09-18
Update dependency libraries and make main.cpp build with clang

jeeb:
Now builds with clang if you specify `NDK_TOOLCHAIN_VERSION=clang`
Also bumps the mindSdkVersion to 21.


app/src/main/jni/main.cpp
```diff
@@ -196,6 +196,9 @@ void sendPropertyUpdateToJava(JNIEnv *env, mpv_event_property *prop) {
         jvalue = env->NewStringUTF(*(const char**)prop->data);
         env->CallStaticVoidMethod(clazz, mid, jprop, jvalue);
         break;
+    default:
+        ALOGV("sendPropertyUpdateToJava: Unknown property update format gotten in callback: %d!", prop->format);
```
joshdekock: replace `gotten` with `received`

PR #29, opened 2016-09-27
Convert the rest of the application to Kotlin

jeeb:
Files left over from the previous conversions were:
- MPVLib
- SummaryEditTextPreference
- MPVFilePickerFragment


app/src/main/java/is/xyz/mpv/MPVFilePickerFragment.kt
```diff
@@ -0,0 +1,44 @@
...
+class MPVFilePickerFragment : FilePickerFragment() {
+
+    override fun setupToolbar(toolbar: Toolbar) {
+        toolbar.visibility = View.GONE
+    }
+
+    override fun onCreateOptionsMenu(menu: Menu, inflater: MenuInflater) {
+    }
+
+    override fun onClickCheckable(v: View, vh: AbstractFilePickerFragment<File>.CheckableViewHolder) = if (!allowMultiple) {
```
xyzz: why is this a lambda?

PR #32, opened 2016-12-10
Remove custom vo/ao from settings

sfan5:


app/src/main/java/is/xyz/mpv/MPVView.kt
```diff
@@ -47,31 +47,20 @@ internal class MPVView(context: Context, attrs: AttributeSet) : GLSurfaceView(co
         val sampleRate = am.getProperty(AudioManager.PROPERTY_OUTPUT_SAMPLE_RATE)
         Log.v(TAG, "Device reports optimal frames per buffer $framesPerBuffer sample rate $sampleRate")
 
-        var ao = "opensles"
+        val ao = "opensles"
```
xyzz: should probably just hardcode them below where it does `MPVLib.setOptionString`

PR #34, opened 2017-01-15
Patch1

nv95:


sfan5:
Thanks for the translation but we'd prefer keeping the icon.

PR #47, opened 2017-06-15
Implement touch gestures

xyzz:
The following are implemented:

* Horizontal swipe to seek
* Vertical swipe on right side to control volume
* Vertical swipe on left side to control brightness

There were some misc changes:
* Volume control stream forced to STREAM_MUSIC (it was unspecified before)
* Immersive mode enabled. This allows us to hide control buttons while detecting touch events properly (Also old `setOnSystemUiVisibilityChangeListener` hack is removed)
* As a result, playback controls now show up on ACTION_UP and only if touch gestures did not kick in.


xyzz:
What I haven't decided yet is what to do with brightness. It starts at 0.5 now, but we probably should remember it somehow.

PR #65, opened 2017-08-30
MPVActivity: Partially implement MXPlayer intent API

### LINKED ISSUE #61: Suport subtitles passed in intent's extra
There are apps which pass URIs for subtitle tracks for the video via extras in the intent.
Currently 2 major APIs seem to be used by video players on android for this functionality (there is no official android/google one)
First one is more popular, but more complex. It is used by MXPlayer and described here:
http://mx.j2inter.com/api (the "subs" object)
Second one is used by VLC, and consists of just and URI inside ~~PLAY_EXTRA_SUBTITLES_LOCATION~~ "subtitles_location" extra.
https://code.videolan.org/videolan/vlc-android/commit/d950a8d6c780e170f98f1722e08e2a522f10df77
I think at least one of them should be implemented in mpv-android, so that it can be used to watch network streams with subtitles.

EDIT: Actually PLAY_EXTRA_SUBTITLES_LOCATION is internal VLC name, the one in intent is subtitles_location https://wiki.videolan.org/Android_Player_Intents/

sfan5:
fixes #61

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
@@ -306,6 +306,8 @@ class MPVActivity : Activity(), EventObserver, TouchGesturesObserver {
         }
     }
 
+    inline private fun <reified T> cast(any: Any?, fallback: T) : T = any as? T ?: fallback
+
     private fun parseIntentExtras(extras: Bundle?) {
         onload_commands.clear()
         if (extras == null)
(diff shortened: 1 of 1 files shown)


jeeb:
Took a look at it and fixed the possibly nullptr casts (at least I think so). Otherwise this seems to be OK, although testing this through `adb` seems rather hard since you cannot seemingly define bytes or arrays of Uris? The furthest I got was: `am start -a "android.intent.action.VIEW" -d "file:///storage/self/primary/Video/HidamariHoneycombED.mp4" --eu "subs" "file:///storage/self/primary/Video/nope.srt" --esn "subs.enable" --ei "decode_mode" 2 -n "is.xyz.mpv/.MPVActivity"`

which results in:
```
09-17 04:40:42.146 28842 28842 W Bundle  : Key decode_mode expected Byte but value was a java.lang.Integer.  The default value 0 was returned.
...
09-17 04:40:42.148 28842 28842 W Bundle  : Key subs expected Parcelable[] but value was a android.net.Uri$StringUri.  The default value <null> was returned.
```
and it doesn't crash, so it seems like nullptrs going into the cast function do not fuck us up.

So @sfan5 - feel free to merge if you wake up before I do if this WorksForYou.

PR #72, opened 2017-10-02
Improve preferences

sfan5:
* prefs: Custom DialogPreference for selecting scaler & its params
* prefs: Allow selecting between CPU/GPU debanding
* prefs: Add preference for selecting video-sync
* prefs: Support for interpolation and tscale filters

screenshot of the custom DialogPreference: (not included: labels moved further right)

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVView.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVView.kt
@@ -88,10 +88,12 @@ internal class MPVView(context: Context, attrs: AttributeSet) : GLSurfaceView(co
                 MPVLib.setOptionString(mpv_option, preference)
         }
 
-        if (sharedPreferences.getBoolean("video_deband", false)) {
-            // use gradfun as --deband=yes did not work on my device's mobile GPUs
-            // also lower the default radius to improve perf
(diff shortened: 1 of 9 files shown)


app/src/main/java/is/xyz/mpv/MPVView.kt
```diff
@@ -85,10 +87,23 @@ internal class MPVView(context: Context, attrs: AttributeSet) : GLSurfaceView(co
...
+            // lower the default radius (16) to improve performance
             MPVLib.setOptionString("vf", "gradfun=radius=12")
+        } else if (deband_mode == "gpu") {
+            MPVLib.setOptionString("deband", "yes")
+        }
+
+        val vidsync = sharedPreferences.getString("video_sync", resources.getString(R.string.pref_video_sync_default))
+        MPVLib.setOptionString("video-sync", vidsync)
+        if (vidsync.startsWith("display-"))
+            MPVLib.setOptionString("opengl-swapinterval", "1")
```
xyzz: isn't that the default?

sfan5: It is indeed, will remove.

Issue #36, opened 2017-01-25
Handle intents for more network streams


scottywz:
Currently, intents are only handled for the `http` scheme.  More schemes, such as https, rtmp, etc., should be handled as well.

sfan5:
Do you have any test pages for when `rtmp`/`rtmps` is passed in an intent?
(Is it just `<a href=...>` like with https?)

andrew-shulgin:
Live RTMP URL example rtmp://tv-penza.ru/live/1

sfan5:
was done in 113b12cf08d0eb9387e10028bae9544efade7b33

PR #75, opened 2017-11-22
Added an intent filter for streaming URL schemes

andrew-shulgin:
Added intent filter for `rtmp`, `rtmpe`, `rtmps`, `rtp`, `rtsp`, `mms`, `mmst`, `mmsh`, and `udp` protocols.
This allows URLs with these protocols to be open directly in MPV.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -21,6 +21,20 @@
             android:launchMode="singleTask"
             android:screenOrientation="sensorLandscape"
             android:theme="@style/AppTheme">
+            <intent-filter>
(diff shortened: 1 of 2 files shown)


sfan5:
fixes #36 btw

are `udp://` URLs ever encountered in the wild?

andrew-shulgin:
Not sure about `udp://`, but mpv should handle it. Why not?

andrew-shulgin:
Just tested opening a multicast `udp://` URL with intent filter. It works.

andrew-shulgin:
UDP streaming is mostly used in IPTV.

jeeb:
Generally LGTM, will be testing as I get a new build going.

PR #76, opened 2017-11-23
Fixed link color

andrew-shulgin:
When clipboard contents with a link were pasted into the **Open URL** dilog, text was almost unreadable (as text color was melding with background color).


sfan5:
Couldn't reproduce bug here, but seems fine to me.

andrew-shulgin:
screenshot_20171123-125842
screenshot_20171123-125901
screenshot_20171123-125908

PR #77, opened 2017-11-23
Toggle seek bar visibility on tap and INFO button press

andrew-shulgin:
Also, reset hide timeout when a touch has been handled.

### DIFF
--- a/app/src/main/java/is/xyz/mpv/TouchGestures.kt
+++ b/app/src/main/java/is/xyz/mpv/TouchGestures.kt
@@ -53,11 +53,11 @@ class TouchGestures(val width: Float, val height: Float, val observer: TouchGest
         trigger = Math.min(width, height) / TRIGGER_RATE
     }
 
-    private fun processMovement(p: PointF) {
+    private fun processMovement(p: PointF): Boolean {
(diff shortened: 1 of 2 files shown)


app/src/main/java/is/xyz/mpv/TouchGestures.kt
```diff
@@ -114,6 +114,6 @@ class TouchGestures(val width: Float, val height: Float, val observer: TouchGest
             }
         }
 
-        return true
+        return false
```
sfan5: doesn't `false` mean the event was not handeled?

andrew-shulgin: You're right. I erroneously thought this return statement is reached only if no gesture has been handled.
I'll fix this.

andrew-shulgin: There are two options:

- Remove onKeyDown from MPVActivity and handle tap with TouchGestures as a PropertyChange;
- Return false and leave the tap handler in MPVActivity.

What do you think is better?

sfan5: latter seems much cleaner to me

PR #80, opened 2017-12-08
Basic playlist support

### LINKED ISSUE #57: Playlist file support
Opening a .pls file currently does not work -- would be quite a useful feature.

sfan5:
not <i>that</i> basic actually
fixes #57

UI:
screenshot_20171208-233157

### DIFF
--- a/app/src/main/res/layout/player.xml
+++ b/app/src/main/res/layout/player.xml
@@ -22,16 +22,14 @@
             android:layout_width="match_parent"
             android:layout_height="match_parent"
             android:orientation="horizontal"
-            android:weightSum="110"
+            android:weightSum="100"
             android:id="@+id/controls_seekbar_group">
 
             <ImageButton
                 android:id="@+id/prevBtn"
                 android:layout_width="wrap_content"
                 android:layout_height="24dp"
-                android:layout_weight="5"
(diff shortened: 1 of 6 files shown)


sfan5:
@Veloxigami if you want to try: https://kitsunemimi.pw/tmp/mpv-android-2017-12-08.apk
`.pls` files are supported, but won't open directly from the browser (use "Open URL")

app/src/main/res/layout/player.xml
```diff
@@ -86,8 +106,6 @@
                 android:layout_height="wrap_content"
                 android:text=".."
                 android:textColor="@android:color/white"
-                android:src="@drawable/ic_short_text_black_24dp"
-                android:tint="#ffffff"
```
sfan5: (this was unused because this is not an `ImageButton`)

qqgg231:
@sfan5 great music taste ^_^

It is not possible to open pls file via file manager (it only shortly blinks), but possible to open via mpv. Is it by design?  Is it possible to fix this?

sfan5:
@qqgg231 when you open via file manager, mpv gets a `content://` URI
mpv-android used to try to resolve it back to the file path, but it no longer does

PR #87, opened 2017-12-23
Copy over the filepicker & fix some stuff

### LINKED ISSUE #7: File picker: when navigating to a directory list Y scroll is not reset
E.g. scroll to the bottom and open "sdcard", instead of displaying ".." at the top like it should, it's scrolled at the same (?) Y pos as the directory you navigated from.

xyzz:
Fixes #7 and #8


xyzz:
apk: http://0x0.st/s8Ml.apk

app/src/main/java/is/xyz/mpv/MPVFilePickerFragment.java
```diff
@@ -1,49 +1,33 @@
...
-        if (viewType != LogicHandler.VIEWTYPE_CHECKABLE)
-            return super.onCreateViewHolder(parent, viewType);
-        View v = LayoutInflater.from(getActivity()).inflate(R.layout.filepicker_item, parent, false);
-        return new CheckableViewHolder(v);
+    public void onClickCheckable(@NonNull View v, @NonNull FileViewHolder vh) {
+        mListener.onFilePicked(vh.file);
     }
 
     public boolean isBackTop() {
         return compareFiles(mCurrentPath, new File("/")) == 0;
```
sfan5: maybe restrict the top path to what was set in settings

sfan5:
[MPL 2.0](https://github.com/spacecowboy/NoNonsense-FilePicker/blob/master/LICENSE) is compatible with MIT I assume

lgtm

app/src/main/java/is/xyz/mpv/MPVFilePickerFragment.java
xyzz: if we do then the user won't be able to navigate to e.g. external sd card (we have /sdcard as default now)

PR #97, opened 2018-02-02
ffmpeg: disable programs, encoders and muxers

myfreeer:
Seldom used on android, disabling these would help reducing build time and binary size.


sfan5:
In theory you can copy the generated programs and use them on-device, so maybe keep those (the savings are small anyway).

sfan5:
62051046529883b605cb8dbc27c75383aaf1ff76

Issue #102, opened 2018-02-22
TODO list for background playback


sfan5:
* ~~app should request audio focus~~ (see #103)
* skip, next/prev buttons
* better notification: duration, song metadata, thumbnail

PR #104, opened 2018-02-23
Background playback improvements

sfan5:
<b>please squash when merging</b>
implements #102 

screen

### DIFF
--- a/app/src/main/jni/thumbnail.cpp
+++ b/app/src/main/jni/thumbnail.cpp
@@ -40,6 +40,7 @@ jni_func(jobject, grabThumbnail, jint dimension) {
             return NULL;
     }
 
+    // extract relevant property data from the node map mpv returns
     unsigned w, h;
     w = h = 0;
     struct mpv_byte_array *data = NULL;
@@ -59,7 +60,8 @@ jni_func(jobject, grabThumbnail, jint dimension) {
             } else if (key == "format") {
                 if (val->format != MPV_FORMAT_STRING)
                     return NULL;
-                if (strcmp(val->u.string, "bgr0"))
+                // check that format equals BGR0
+                if (strcmp(val->u.string, "bgr0") != 0)
                     return NULL;
             } else if (key == "data") {
                 if (val->format != MPV_FORMAT_BYTE_ARRAY)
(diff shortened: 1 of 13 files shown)


xyzz:
You need to use delete[] instead of delete.

app/src/main/jni/thumbnail.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+    mpv_free_node_contents(&result); // frees data->data
+
+    // convert & scale to appropriate size
+    struct SwsContext *ctx = sws_getContext(
+        new_w, new_h, AV_PIX_FMT_BGR0,
+        dimension, dimension, AV_PIX_FMT_RGB32,
+        SWS_BICUBIC, NULL, NULL, NULL);
+    if (!ctx)
+        return NULL;
```
xyzz: You're leaking new_data here. Probably better to replace with std::vector or similar so that RAII can be utilized.

app/src/main/jni/thumbnail.cpp
```diff
@@ -0,0 +1,131 @@
...
+        crop_top = tmp / 2;
+        crop_bottom = tmp / 2 + tmp % 2;
+    }
+    unsigned new_w, new_h;
+    new_w = w - crop_left - crop_right;
+    new_h = h - crop_top - crop_bottom;
+    ALOGV("cropped w:%u h:%u\n", new_w, new_h);
+
+    std::vector<uint32_t> new_data;
+    new_data.reserve(new_w * new_h);
```
xyzz: You can't use "new" vector members after you call `reserve`, you need to do `resize`, or just pass the size as argument to the constructor.

app/src/main/jni/thumbnail.cpp
```diff
@@ -0,0 +1,131 @@
...
+        new_w, new_h, AV_PIX_FMT_BGR0,
+        dimension, dimension, AV_PIX_FMT_RGB32,
+        SWS_BICUBIC, NULL, NULL, NULL);
+    if (!ctx)
+        return NULL;
+    int src_stride = sizeof(uint32_t) * new_w, dst_stride = sizeof(uint32_t) * dimension;
+    std::vector<uint8_t> scaled;
+    scaled.reserve(dimension * dst_stride);
+    uint8_t *src_p[] = { (uint8_t*) new_data.data() }, *dst_p[] = { scaled.data() };
+    sws_scale(ctx, src_p, &src_stride, 0, new_h, dst_p, &dst_stride);
```
xyzz: All arrays you pass to `sws_scale` have to be 4 elements, it's a bug/undocumented in ffmpeg, so make sure src/dst/stride have 4 elements.

app/src/main/jni/thumbnail.cpp
```diff
@@ -0,0 +1,131 @@
...
+    }
+    unsigned new_w, new_h;
+    new_w = w - crop_left - crop_right;
+    new_h = h - crop_top - crop_bottom;
+    ALOGV("cropped w:%u h:%u\n", new_w, new_h);
+
+    std::vector<uint32_t> new_data;
+    new_data.reserve(new_w * new_h);
+    for (int y = 0; y < new_h; y++) {
+        for (int x = 0; x < new_w; x++) {
```
xyzz: can the inner loop be replaced with memcpy?

jeeb:
Merged as 6a7cd9eee8d8e6101ab31f94f2d09fc1022bedba

PR #113, opened 2018-03-25
Audio preferences with audio delay

junaru:
Allows user to set audio-delay via the settings menu. 
Main reason for implementing this was to have a way to compensate for sound delay on bluetooth headset when watching video with hardware decoding enabled.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -63,7 +63,7 @@
     <string name="pref_header_audio">Audio</string>
 
     <string name="pref_audio_delay_title">Audio delay</string>
-    <string name="pref_audio_delay_summary">Usefull to compensate for out of sync audio.</string>
+    <string name="pref_audio_delay_summary">Useful to compensate for out of sync audio.</string>
 
     <string name="pref_audio_delay_offset_title">Audio delay offset</string>
     <string name="pref_audio_delay_offset_hint">Offset in seconds (+/-)</string>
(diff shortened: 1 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -60,6 +60,14 @@
     <string name="pref_video_fastdecode_title">Low-quality video decoding</string>
     <string name="pref_video_fastdecode_summary">Trades quality for speed and thus fluent playback.\nREDUCES QUALITY A LOT</string>
 
+    <string name="pref_header_audio">Audio</string>
+
+    <string name="pref_audio_delay_title">Audio delay</string>
+    <string name="pref_audio_delay_summary">Usefull to compensate for out of sync audio.</string>
```
sfan5: `Useful`*

sfan5:
Since this can also be done by editing mpv.conf, I think this PR would provide more value if it used a slider or similar for easier tuning.

junaru:
Indeed slider would look nicer but we would lose the ability to fine tune the setting.
If I'm not mistaken most players do audio delay in 10ms increments and for example my Bluetooth headphones introduce about 730ms of it.  
So slider like control would have several issues:

1. A need for 'reasonable' hard coded interval and that alone will probably ruin someones day.
2. IMHO 'reasonable' would be [-3000ms:+3000ms], given 10ms increments that's 600 distinct values that you have to fit on a slider that would probably be more frustrating to use than the current edit box.
3. A number picker like widget could have a bigger interval but user would still need to scroll through all the 10ms increments. 

This is a fringe feature that not many would probably use to begin with and since mpv.conf editing was merged no hard feelings would be had if this was rejected altogether.

I'd personally like it to stay as number input.

sfan5:
closed in favour of editing mpv.conf

PR #126, opened 2018-06-12
Playlist support, implicit playlists

xyzz:
This adds initial playlist support and implicit playlists. Implicit playlist means that when you click on a file in our file picker, instead of playing that one file it creates a playlist of all files in the directory and plays current file.


sfan5:
works nicely as far as I've tested, just two things:
* the UI no longer fades out but just disappears at some point
* the background playback stuff (`BackgroundPlaybackService.kt`) is not aware of the playlist

nikhilCad:
What happened here. Its been two years, any reason this is on hold?

PR #134, opened 2018-07-12
Handle save-position-on-quit manually instead of relying on mpv

sfan5:
doesn't work correctly otherwise

TODO<i>?</i>: if you kill mpv from the app drawer, the position will not be saved
it's possible to solve this but the solution is hacky/not trivial

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
@@ -68,6 +68,8 @@ class MPVActivity : Activity(), EventObserver, TouchGesturesObserver {
 
     private var backgroundPlayMode = ""
 
+    private var shouldSavePosition = false
+
     private fun initListeners() {
         controls.cycleAudioBtn.setOnClickListener { _ ->  cycleAudio() }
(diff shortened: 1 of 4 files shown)


philipl:
Even without handling killing from the app drawer, this is hugely useful. Thanks for looking into it. Change looks reasonable to me for whatever that's worth.

seowalex:
I've done some experimenting with lua scripting and from what I can see wouldn't the watch later config not get deleted on eof?

sfan5:
The code makes sure to only write the watch later config when you are actually exiting the app (also not on EOF).
Nothing to delete in that case.

seowalex:
I understand, but let's say you exit the app once. This will result in a watch later config being created. After than, if you continue watching the video until the end, wouldn't the old watch later config still persist?

rieje:
When I hit back and then open the video again, it still resumes the video without the video (i.e. it resumes with just the audio). I'm on the latest version (2018-04-29-alpha / 14).

philipl:
There hasn't been a new official build since this change went in, so that's not a surprise. I did my own build and confirmed that it works.

PR #143, opened 2018-10-01
Add a button for cycling through playback speeds

KoffeinFlummi:
Cycles through 0.1x, 0.5x, 0.75x, 1x, 1.25x, 1.5x, and 2x. I tried setting speed via dragging, but that made it difficult to reset to 1x.

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVView.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVView.kt
@@ -220,6 +220,10 @@ internal class MPVView(context: Context, attrs: AttributeSet) : SurfaceView(cont
     val hwdecActive: Boolean?
         get() = MPVLib.getPropertyString("hwdec-current") != "no"
 
+    var playbackSpeed: String?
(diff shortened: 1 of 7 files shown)


app/src/main/java/is/xyz/mpv/MPVView.kt
```diff
@@ -220,6 +220,10 @@ internal class MPVView(context: Context, attrs: AttributeSet) : SurfaceView(cont
     val hwdecActive: Boolean?
         get() = MPVLib.getPropertyString("hwdec-current") != "no"
 
+    var playbackSpeed: String?
+        get() = MPVLib.getPropertyString("speed").toFloat().toString()
```
xyzz: Why do you convert a string to float then back to string again? Also, shouldn't the property be a float?

KoffeinFlummi: There were no property methods for double properties, and the reconversion was for getting rid of excessive zeroes. I added those methods now.

PR #150, opened 2018-10-29
Automatic orientation depending on video aspect ratio

sfan5:
perhaps the most requested feature in play store reviews


sfan5:
You mean instead of the `Force (Landscape|Portrait)` options?
Since IMO the auto mode is the most useful part of this feature.

sfan5:
When using a setting other than "Automatic", this thingy will appear and do exactly what it looks like:
screen

xyzz:
Okay this looks very good but is there any reason not to show the button all the time?

sfan5:
Making sure the automatic mode does (or doesn't) do its thing when the user can also manually change the orientation is somewhat annoying due to events. Which is why I haven't bothered.

PR #152, opened 2018-11-05
remove no longer relevant sentence in buildscripts/README.md

theqp:
clang is default, gcc can be enabled instead, however it is unsupported


jeeb:
Merged in as 5b901d8977f8879a78ad7ba740f794e8d83b96b9 after editing the commit message.

PR #179, opened 2019-04-22
Support additional Arabic and Hebrew characters

LukePulverenti:
This contains fixes for these two languages based on issues reported to us (Emby) over the last year.


sfan5:
Where does this font file come from?

LukePulverenti:
From this repository. We simply added additional characters.

sfan5:
Same method as #56 then, I guess

PR #187, opened 2019-05-31
Add optional one-way command server

MarkMendell:
Adds a preference to enable a server listening on abstract socket "is.xyz.android" that reads lines from its clients and runs `mpv_command_string(line)` (like the `--input-ipc-server` option in `mpv`).

**Security note:** since it's a socket, there's no restriction on which apps get to pass arbitrary text to `mpv_command_string`. The setting warns about this. I'm not familiar with other Android IPC but am happy to learn and change it.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -21,6 +21,7 @@ android {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
+    implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.2.1'
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
     implementation 'androidx.appcompat:appcompat:1.0.2'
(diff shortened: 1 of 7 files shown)


sfan5:
This would've been much simpler to implement using intents like [`NotificationButtonReceiver`](https://github.com/mpv-android/mpv-android/blob/master/app/src/main/java/is/xyz/mpv/NotificationButtonReceiver.kt) does for the notification buttons.

xyzz:
What's the use case?

MarkMendell:
> This would've been much simpler to implement using intents like [`NotificationButtonReceiver`](https://github.com/mpv-android/mpv-android/blob/master/app/src/main/java/is/xyz/mpv/NotificationButtonReceiver.kt) does for the notification buttons.

Agreed. I'll open up a new PR and close this one.

> What's the use case?

I run mpv in split-screen with [Termux](https://termux.com/). On Termux I have a shell script that advances after each subtitle once I understand it (I'm learning Korean).

PR #195, opened 2019-07-28
Add Chinese translations for mpv-android

giwhub:


sfan5:
The last few commits had some translation-relevant changes. Would you mind rebasing your commits?
Also you don't need to copy the values marked `translatable="false"`.

giwhub:
Okay, I would pull another request, closed.

PR #208, opened 2019-09-29
Minor UI & available speed modes changes

gmikhail:
Аннотация 2019-09-29 170727


sfan5:
The subtitle icon looks too much like "volume" to me, I prefer the old one.

gmikhail:
How about this?
https://material.io/resources/icons/?search=audiotrack&icon=audiotrack&style=baseline

PR #291, opened 2020-07-30
add basic dpad navigation support (AndroidTV)

mariasanchez84:
- add basic dpad navigation support for Android TV

DPAD_UP and DOWN both control activation and deactivation of the dpad navigation with the interceptDpad function.
DPAD_LEFT, RIGHT and CENTER handle the button selection and trigger, but they are only overridden when dpad navigation is enabled.
Also the callback to auto hide the controls is disabled as long as dpad navigation is active, only.

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
@@ -58,8 +58,8 @@ class MPVActivity : Activity(), MPVLib.EventObserver, TouchGesturesObserver {
     private lateinit var audioManager: AudioManager
 
     private var btnSelected = -1
-    private val colorFocussed = Color.argb(128, 128, 128, 176)
-    private val colorNoFocus = Color.argb(0, 0, 0, 0)
+    private val colorFocussed = ContextCompat.getColor(applicationContext, R.color.tint_btn_bg_focussed)
+    private val colorNoFocus = ContextCompat.getColor(applicationContext, R.color.tint_btn_bg_nofocus)
 
     private val seekBarChangeListener = object : SeekBar.OnSeekBarChangeListener {
         override fun onProgressChanged(seekBar: SeekBar, progress: Int, fromUser: Boolean) {
@@ -129,19 +129,6 @@ class MPVActivity : Activity(), MPVLib.EventObserver, TouchGesturesObserver {
 
         controls.prevBtn.setOnLongClickListener { pickPlaylist(); true }
         controls.nextBtn.setOnLongClickListener { pickPlaylist(); true }
-    }
-
-    private fun dpadCenterTriggerAction(btnId: Int) {
-        when (btnId) {
-            controls.playBtn.id -> playPause(controls)
-            controls.cycleAudioBtn.id -> cycleAudio()
-            controls.cycleSubsBtn.id -> cycleSub()
-            controls.cycleDecoderBtn.id -> switchDecoder(controls)
-            controls.cycleSpeedBtn.id -> cycleSpeed(controls)
-            controls.menuBtn.id -> openTopMenu(controls)
-            controls.nextBtn.id -> playlistNext(controls)
-            controls.prevBtn.id -> playlistPrev(controls)
-        }
     }
 
     @SuppressLint("ShowToast")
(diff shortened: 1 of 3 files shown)


app/src/main/java/is/xyz/mpv/MPVActivity.kt
```diff
@@ -56,6 +57,10 @@ class MPVActivity : Activity(), MPVLib.EventObserver, TouchGesturesObserver {
     private lateinit var gestures: TouchGestures
     private lateinit var audioManager: AudioManager
 
+    private var btnSelected = -1
+    private val colorFocussed = Color.argb(128, 128, 128, 176)
```
sfan5, line 61: should probably use a tint like `updatePlaylistButtons` does

app/src/main/java/is/xyz/mpv/MPVActivity.kt
```diff
@@ -126,6 +131,19 @@ class MPVActivity : Activity(), MPVLib.EventObserver, TouchGesturesObserver {
         controls.nextBtn.setOnLongClickListener { pickPlaylist(); true }
     }
 
+    private fun dpadCenterTriggerAction(btnId: Int) {
+        when (btnId) {
```
sfan5, line 135: This should definitely call `onClick()` (or whatever the name is) of the selected view directly.
That'd make the when statement unnecessary and also be "portable" for future changes.

app/src/main/java/is/xyz/mpv/MPVActivity.kt
```diff
@@ -506,6 +527,64 @@ class MPVActivity : Activity(), MPVLib.EventObserver, TouchGesturesObserver {
...
+                    }
+                }
+                return true
+            }
+        }
+        return false
+    }
+
+    fun updateShowBtnSelected () {
+        if (controls_button_group != null) {
```
sfan5, line 574: `controls_button_group` will never be null

app/src/main/res/layout/player.xml
```diff
@@ -186,6 +197,7 @@
         android:textStyle="bold"
         android:visibility="gone" />
 
+    <!--
```
sfan5, line 200: The top controls should definitely be kept.
I suggest making the menu button selectable by reserving e.g. the last index for it. Then it doesn't need to be moved to the control group.

sfan5:
The usecase for this is Android TV, right?

mariasanchez84:
Yep, I've been using this on my android TV box (...with a very minimalist remote).

app/src/main/java/is/xyz/mpv/MPVActivity.kt
mariasanchez84: Done.
`val colorFocused = ContextCompat.getColor(applicationContext, R.color.tint_btn_bg_focused)`

app/src/main/java/is/xyz/mpv/MPVActivity.kt
mariasanchez84: Way better indeed.
`controls_button_group.getChildAt(btnSelected)?.performClick()`

app/src/main/java/is/xyz/mpv/MPVActivity.kt
mariasanchez84: Removed.

app/src/main/res/layout/player.xml
mariasanchez84: Done.
Looked good in the control bar though, but it's probably better to have a separate menu.

mariasanchez84:
Everything should be fixed now.
Also, I hardcoded these for conveniency if you don't mind :
```
// AndroidTV
KeyEvent.KEYCODE_MENU -> openTopMenu(controls)
KeyEvent.KEYCODE_GUIDE -> openTopMenu(controls)
KeyEvent.KEYCODE_ENTER -> player.cyclePause()
KeyEvent.KEYCODE_DPAD_CENTER -> player.cyclePause()
```

sfan5:
Works fine, thanks for contributing this feature.

sfan5:
Merged as 7ec2052b24f221d62c4692c4b3823f305e3bb479

PR #296, opened 2020-08-26
MPVActivity: Add pref to extend video in the cutout area

philipl:
Many newer Android devices feature a notch or hole-punch camera, and by
default, Android will size landscape windows so that they don't extend
into the cutout area. For video playback, it is common to support
displaying the video over the entire screen, regardless of the cutout,
so let's add that option.

I've defaulted this to off, but maybe it should be on.


philipl:
Updated
* do not assume API level
* moved pref to General
* defaulted to on.

sfan5:
Merged 1a8505c7cbd8e59d68312a65592d6a9609a591ee

jackloomen:
This appears to not work on Android 12, Pixel.

PR #314, opened 2020-09-30
Updated links in README.md

bluecypher:
Thanks for the app!

You can add buttons for the F-Droid and Google Play links to your app.


sfan5:
Please squash the commits into a single one and give it a meaningful commit message.

PR #334, opened 2020-11-17
TouchGestures: Add a gesture to toggle Pan/Scan zooming

philipl:
Most modern Android phones have very wide screens and media players
often offer the ability to zoom to remove black bars. mpv's panscan
property does this, so let's add a gesture to allow toggling it.

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
@@ -1478,6 +1478,8 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
                 fadeGestureText()
             }
             PropertyChange.PlayPause -> player.cyclePause()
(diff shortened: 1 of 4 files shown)


sfan5:
This works fine (minus small code style issues), but I wonder if gestures are the right place for this functionality.

I think the top menu is a better place for a panscan slider/toggle, but that menu currently suffers from not being able to put more into it and I'd need to get around to redesigning it first.

philipl:
It seemed convenient to make it a gesture, and sort of analogous to how the youtube app does it (albeit that is a pinch gesture).

sfan5:
I decided that it's good to add this after all.
Two changes are needed however:
* MPVActivity shouldn't store state for this, use `cycle-values panscan 1.0 0.0`
* The array in `res/values-ja/arrays.xml` needs to be adjusted too (just use the English text)

philipl:
Done.

PR #346, opened 2021-01-07
add Japanese translation

ray4423:


app/src/main/res/values-ja/arrays.xml
```diff
@@ -0,0 +1,42 @@
...
+<resources>
+    <string-array name="background_play_entries">
+        <item>しない</item>
+        <item>オーディオファイルのみ</item>
+        <item>いつも(ビデオファイルも含む)</item>
+    </string-array>
+    <string-array name="auto_rotation_entries">
+        <item>自動</item>
+        <item>ランドスケープ</item>
+        <item>ポートレート</item>
```
sfan5, line 11: Google uses 横向き and 縦向き for Landscape / Portrait in their documentation. I think those would be more accurate here.

app/src/main/res/values-ja/strings.xml
```diff
@@ -0,0 +1,143 @@
...
+    <string name="pref_hardware_decoding_summary">ソフトウェアデコードの代わりに(利用可能なら)ハードウェアデコードを使用します。</string>
+
+    <string name="pref_background_play_title">バックグラウンド再生</string>
+    <string name="pref_background_play_summary">アプリがバックグランドになった場合にそのまま再生する/停止するの設定</string>
+
+    <string name="pref_save_position_title">終了時に再生位置を保存</string>
+    <string name="pref_save_position_summary">各ファイルの再生終了時に停止位置を保存します。同じファイルを再生した際に前回終了時の位置から再開します。</string>
+
+    <string name="pref_auto_rotation_title">画面の方向</string>
+    <string name="pref_auto_rotation_summary">再生時にランドスケープ/ポートレートにするかの設定</string>
```
sfan5, line 79: (also mentioned here)

app/src/main/res/values-ja/arrays.xml
ray4423: I agree.
I translated it as "ランドスケープ" with reference to vlc, 
but it is better to match it with google manners.

Should I pull request the modified one again?

ray4423: I pushed the fixed one instead of pull request again.
Is it OK?

sfan5: Sure, thanks.

PR #425, opened 2021-06-29
Added sftp protocol scheme

0xACE:
This change should allow external tools who are prompting for sftp:// to
find mpv as a potential target.

This specific use case has appeared in MiXplorer where it prompted for
sftp:// and later shares a http:// link to the actual content.

I haven't tested if this is the problem because I can't build it on my machine atm. But I strongly suspect that this is the issue.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -69,6 +69,8 @@
 
                 <data android:scheme="http" />
                 <data android:scheme="https" />
+                <data android:scheme="ftp" />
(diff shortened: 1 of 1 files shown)


0xACE:
I talked to the developer of MiXplorer. He couldn't replicate this issue, in fact he could get `mpv-android` listed. I realised it may be my environment that is the problem.

I uninstalled MX Player, and MiXplorer is now giving many other options for opening media files.

Seems like this PR is not necessary, but I'm leaving this comment here in case anyone runs into the same issue.

Issue #203, opened 2019-09-07
Seek gestures conflict with android 10 navigation


philipl:
The new android 10 gesture navigation conflicts fairly badly with the seek gestures. You have to use 'back' to return from the player to the file browser and you're much more likely to seek in the player than manage a back gesture. Seems like we need an explicit area where seek won't trigger, like having seek in the top half and back in the bottom half, for example.

philipl:
Given that the more likely conflict is seeking when you want to go back, adding a stop button would avoid the need to use the back gesture. That could be done as configuration via #206 if not a first class action.

KaKi87:
Hello,

I'm experiencing the same issue.

For now I can't find any other way than killing and re-opening the app.

Any news on the button idea ?

Thanks

sfan5:
You now need to show the player UI first before performing a gesture.

KaKi87:
You mean click then swipe instead of just swipe ? That's really bad UX.

sfan5:
Yes.
Bad UX for a video player that's usually in immersive mode anyway? Hmm I wouldn't say so.

KaKi87:
Before this change, seeking would require : touchstart -> touchmove -> touchend.

After this change, seeking requires : touchstart -> touchend -> touchstart -> touchmove -> touchend.

PR #466, opened 2021-12-08
Horizontal deadzone to avoid conflict with A10+ Back gesture.

Long-FA:
If Hold to seek is enabled, attempting to exit fullscreen mode with Android 10+ Back gesture will result in MPV seeking the video. As a workaround, users may first swipe from top/bottom to exit fullscreen mode, then swipe from left/right side for back gesture, but it's definitely not as intuitive as double swiping back gesture. Similar to a107eff.


sfan5:
relevant issue: #203
Not sure I like the extra deadzone, but I don't have better idea.

PR #469, opened 2021-12-15
also include total duration in result intent

jmir1:
I'd like to integrate mpv into my project and kinda need to check if the user has finished watching the video.
in order to do that, i need the total duration.
mx player uses the same extra key for the duration.


sfan5:
Hmmm. Sounds okay but wouldn't the calling application know the entire length of the file anyway?

jmir1:
not if it's just a URL to an online video.

Issue #517, opened 2022-04-05
can we not build arm 32bit and split APKs?


Quackdoc:
It would be nice to skip building arm 32bit support. I am currently working on getting github CI support for building MPV-Android and it seems to be working well for the yt-dl branch as well apk signing works too.

however I would like to decrease the run time of the builds. currently the total run time is about 15-20 min for x86_64 and arm64, and 9-11 min for just arm32. I'm not expecting a 10min improvement or anything, (already decreased the run time from 30 min to 20 min by building x64 and arm64 separately), however I would like to cut the time down a tad bit more if at all possible. 

EDIT: as an appendage, would it be possible to set it up to split the APKs? seems like it would be preferred for fdroid.

https://gitlab.com/fdroid/fdroiddata/-/issues/2643

linsui:
@sfan5 This scheme of the version code doesn't work well with F-Droid because F-Droid only keep the apks with the highest version code in the main repo. So after some releases only the x86-64 apks are kept. Could you please use the abi codes as suffix instead so that newer version always have higher version code? Thanks!

sfan5:
The version scheme is the [official recommendation from Google](https://developer.android.com/studio/build/configure-apk-splits#configure-APK-versions). I could probably swap the places of the two components.
But surely this isn't the only project which does this, maybe f-droid can just patch it out?

linsui:
This is the second app I have seen with this scheme though. I thought it's not a good idea to allow the user update to an old version.
Yes, we can patch it. I would like to keep the same version code with your release but it's not a big problem.  :ok_hand:

linsui:
I opened an [issue](https://gitlab.com/fdroid/fdroidserver/-/issues/1011) to track this problem in F-Droid.

sfan5:
Thanks, I'm interested to know what comes of it.

PR #534, opened 2022-05-12
Reduce app size

TyraVex:
It works perfectly for me, but I don't know about other devices.
I think the subfont replacement is something that can be easily merge, the rest is more experimental

### DIFF
--- a/buildscripts/scripts/ffmpeg.sh
+++ b/buildscripts/scripts/ffmpeg.sh
@@ -23,8 +23,8 @@ cpuflags=
 [[ "$ndk_triple" == "arm"* ]] && cpuflags="$cpuflags -mfpu=neon -mcpu=cortex-a8"
 
 ../configure \
-	--target-os=android --cross-prefix=$ndk_triple- --cc=$CC \
-	--arch=${ndk_triple%%-*} --cpu=$cpu --enable-{jni,mediacodec,libdav1d} \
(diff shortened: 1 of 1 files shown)


sfan5:
**1**: The shipped font is the only one that mpv can make use of (\*) so it needs to have as wide coverage as possible. Your font file barely covers more than [Latin-1](https://de.wikipedia.org/wiki/ISO_8859-1), which is okay if you only watch with English (or no) subtitles but this doesn't cut it for everyone.
(\*) Unless the media file comes with embedded subtitle fonts, plain subtitles don't have this.

**2**: Half of the options you added to ffmpeg don't do anything. Some of the rest are bad for performance, some break certain formats in subtle ways. Support for HTTP or any streaming formats is gone too.
What probably makes the largest difference (without any side effects) is `--disable-filters`. There are a *lot* of filters in FFmpeg and 98% of users will never be using any but for the people who do it's hard to throw them away...

What would really help reduce the APK/download size without any downsides is split APKs #517, which I plan to do next release.

TyraVex:
Okay, thanks for the explanations.
I think it is okay to keep the filters for the 2% who uses them (even if I don't understand how they are useful while playing videos tbh).
Because at the end, the app is still less than 50mb, so  it's not that big of a deal considering its capabilities to play such a variety of video formats.

TyraVex:
And sorry for this not very helpful PR, it's my first one.  But at the end I'm glad I learned a bit more about git and mpv/ffmpeg :)

Issue #599, opened 2022-09-14
Settings do not take effect after updating to Android 12


h0ld0:
Android version: Android 12

mpv-android version:
2022-7-30-release

Which version of mpv-android introduced the problem (if known):
2022-7-30-release

Description
After updating my Galaxy a12 to android 12, settings like remember choice on next startup, save position on quit, display media title, and perhaps a couple more doesn't take effect. it may be caused by a tightened security update on the OS. reinstalled and cleared cache/data yields no difference.

sfan5:
But you can see the values you set in the settings? And if you set them again it makes no difference?

h0ld0:
> But you can see the values you set in the settings? And if you set them again it makes no difference?

yes. they're toggled on, but it doesn't seem to take effect. also, the UI randomly generates 3 digits as their media title when set to display their file names. 

![Screenshot_20220914-211102_mpv](https://user-images.githubusercontent.com/113484063/190163397-9818ab5b-92de-4a27-9d96-880ce48b1f5c.jpg)
![Screenshot_20220914-210735_mpv](https://user-images.githubusercontent.com/113484063/190163413-aecbe0d1-c321-4a3e-a916-42f370730644.jpg)

YanceyChiew:
I believe this is related to issue #542. When the current version of mpv-android opens a content uri, what is passed to libmpv is just a file descriptor, or put another way, a number that has nothing to do with the filename. So, of course all settings related to filename or path will be invalid.

sfan5:
> yes. they're toggled on, but it doesn't seem to take effect.

In your screenshots it's clearly showing that the "show media title" option is taking effect. This doesn't fit with your bug report.

Issue #578, opened 2022-07-31
Incorrect file names are displayed


CptnLevi69:
In the latest version of mpv, instead of displaying the file name specified in the video file itself, it just shows a random number in place of the file name. 

Due to this, it's also really hard to organize/distinguish between screenshots as the screenshot file is just named as that number and not the file name.

sfan5:
fundamentally the same issue as #542, see the explanations there
use the legacy file picker if it works for you.

CptnLevi69:
Using the legacy file picker resolves the issue completely so I assume it has to do with how the document picker handles files?

Ultra-Code:
How do you set the legacy file picker as the default? I am opening the video file from telegram and not directly from mpv.

sfan5:
If you don't open the video from mpv-android's file picker this advice does not apply.

Ultra-Code:
For my case is there any suggestion?

BrendonIrwan:
Hi all. It appears to be fixed with the latest build. Tested and confirmed on two devices.

Issue #542, opened 2022-05-28
Videos played from content uri stores and loads wrong progress position save file


YanceyChiew:
### Important information

Android version: 12

mpv-android version: 2022-03-23 release

### Description

When loading a video document from a content provider, I often encounter situations where the player quits immediately or cannot properly read valid data from the provider because the current position exceeds the total video duration.

Or sometimes the video player automatically quits in the middle of playing, and it doesn't look like it crashes when it encounters an error. The postion when quitting is always as long as the total duration of the previous video.

- Use the following commands to generate 3 video files for testing

```bash
mkdir /sdcard/test
cd test
ffmpeg -f lavfi -i nullsrc=s=256x256:d=1800 test30min.mp4
ffmpeg -f lavfi -i nullsrc=s=256x256:d=600 test10min.mp4
ffmpeg -f lavfi -i nullsrc=s=256x256:d=180 test3min.mp4
```
Opening them through a traditional file browser, mpv handles the saving progress position normally, however, when opened by the content provider _(ie. `com.google.android.documentsui`)_ , mpv's handling of the save position is confusing.

### Log output

Since there is too much output, I only filter key messages, and mpv doesn't output the open watch_later filename every time it plays.     _mpv has uid 10667 in my device_

	logcat -c && logcat --uid 10667|grep -C2 -E "Resolving|watch_later|'start'|Exiting"

I opened the 10-minute version first, dragged the progress to the end, exited and opened the 30-minute version, then exited and opened the 3-minute version, and finally reopened the 10-minute version.

- Below is the output
```
ar && logcat --uid 10667|grep -C1 -E "Resolving|watch_later|'start'|Exiting"                                    <
05-29 04:34:27.211 17273 17273 E Toast   : setGravity() shouldn't be called on text toasts, the values won't be used
05-29 04:34:27.212 17273 17273 V mpv     : Resolving content URI: content://com.android.externalstorage.documents/document/primary%3Atest%2Ftest10min.mp4
05-29 04:34:27.239 17273 17273 V mpv     : Display 0 reports FPS of 60.000004
--
05-29 04:34:32.033 17273 19054 V mpv     : [cplayer:v] Set property: vo="null" -> 1
05-29 04:34:32.042 17273 17273 V mpv     : Exiting.
05-29 04:34:37.432 17273 17688 D AppScoutStateMachine: 17273-ScoutStateMachinecreated
--
[...]

PR #601, opened 2022-09-20
MPVActivity: Pass content uri to libmpv.

YanceyChiew:
Pass content uri to libmpv via the fdclose protocol, which will treat it as a path.

Do not merge until [this](https://github.com/mpv-player/mpv/pull/10659) upstream pr has passed.

close:
  - #542
  - #578
  - #599

_Known issue: Remember playback position doesn't work with mpv-android's built-in document picker, but still displays titles correctly._

_Here is an [artifact](https://github.com/YanceyChiew/mpv-andorid-temp/actions/runs/3086478625) available for temporary testing._

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
@@ -909,7 +909,7 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
             }
         } catch(e: Exception) { }
         // Else, pass the fd to mpv
-        return "fdclose://${fd}"
+        return "fdclose://${fd}/p/${uri}"
     }
 
     private fun parseIntentExtras(extras: Bundle?) {


YanceyChiew:
> Known issue: Remember playback position doesn't work with mpv-android's built-in document picker, but still displays titles correctly.

This is because libmpv uses the original string file name in the playlist when saving the watch_later file, i.e. the unprocessed url, which of course changes with the fd.

For other document providers, the file descriptors provided in a short period of time may be the same, so it creates the illusion that the playback position can be recorded correctly. In fact, this problem has not been solved.

---

Since the `ignore-path-in-watch-later-config` option is not valid for the url, I think I might eventually need to hook `on_load` to modify the `stream-open-filename` property to solve the problem.

sfan5:
I am not convinced that this is the fix we want. As you said yourself URLs are still unpredictable due to the fd component so the watch later feature will not work properly.
In the end we'll need a hook that resolves `content://` URLs on demand and that hook can probably also fix the title.

However thinking further hooks can't apply to stuff like external files or additional audio/subtitle tracks. So the only real fix here are [stream callbacks](https://github.com/mpv-player/mpv/blob/master/libmpv/stream_cb.h) (plus something custom for media titles)...

YanceyChiew:
> However thinking further hooks can't apply to stuff like external files or additional audio/subtitle tracks. So the only real fix here are [stream callbacks](https://github.com/mpv-player/mpv/blob/master/libmpv/stream_cb.h) (plus something custom for media titles)...

This is indeed the optimal solution. I didn't realize that mpv opened such an API before.

In fact, as long as the `stream_cb` of the content protocol is implemented, the file title can be displayed correctly without additional processing, because after testing, I found that the file name and path in the stream structure are ignored, and what actually works is only the `MPContext->filename`, which came from the original file name in the playlist. 

Because there are strict restrictions on the format of the fd and fdclose protocols in the code of mpv, no other characters are allowed after the file descriptor, so this pr cannot be used alone as a temporary solution, I will close it after verifying some related issues.

PR #602, opened 2022-09-21
no exit_warning_playlist if save_position

louis-19:


app/src/main/java/is/xyz/mpv/MPVActivity.kt
```diff
@@ -774,6 +774,12 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
         if (lockedUI)
             return showUnlockControls()
 
+        val prefs = getDefaultSharedPreferences(applicationContext)
+        if (prefs.getBoolean("save_position", false)) {
+            finishWithResult(RESULT_OK, true)
+            return
+        }
```
sfan5, line 781: you can directly retrieve the settings from `this.shouldSavePosition`
and also it'd be better to integrate it with if-clause that already exists

louis-19: Thank you for the comment!
Is there any other reason to give the exit warning? Maybe better off to have a user config option to disable it.

sfan5: The warning is so you don't lose your playlist, I don't think there's a reason to disable it right now.

louis-19: That's my second thought after the PR. Then this PR is irrelevant as my use case is directory playlist.

PR #637, opened 2022-12-24
Build fdk-aac

Quackdoc:
My main use for this was to build my own fdk-aac for support for other aac codecs, but libfdk-aac has good performance anyways. 

I have it set to only use gnu compatible fdk-aac (at least according to RHEL anyways). So while `--non-free` is toggled, it should still only be compiling against gnu compatible and should be fine for distribution.

I'm not sure if merely toggling `--non-free` makes ffmpeg unlicensed.

### DIFF
--- a/buildscripts/include/depinfo.sh
+++ b/buildscripts/include/depinfo.sh
@@ -21,7 +21,8 @@ v_mbedtls=2.28.1
 
 dep_mbedtls=()
 dep_dav1d=()
-dep_ffmpeg=(mbedtls dav1d)
+dep_fdkaac=()
+dep_ffmpeg=(mbedtls dav1d fdkaac)
 dep_freetype2=()
 dep_fribidi=()
 dep_harfbuzz=()
--- a/buildscripts/include/download-deps.sh
+++ b/buildscripts/include/download-deps.sh
@@ -50,6 +50,13 @@ if [ ! -d lua ]; then
 		tar -xz -C lua --strip-components=1
 fi
 
+#fdkaac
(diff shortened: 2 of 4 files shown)


sfan5:
FFmpeg supports decoding HE-AAC and HE-AACv2 so not sure what your goal is here. xHE-AAC?

Quackdoc:
that's correct, it's not something that can be distributed sadly, but allegedly FDK-AAC also preforms better then ffmpeg's AAC implementation, I'm not sure if that applies for decoding or encoding. 

however I figured if the chance is there even though fdkaac-free doesn't have xHE-AAC if it's true that it preforms better, and I have the stuff plumbed up anyways I may as well make a PR anyways. plus this makes it simple for individuals to make their own builds by simply swapping the fdkaac thats downloaded

sfan5:
> that's correct, it's not something that can be distributed sadly, but allegedly FDK-AAC also preforms better then ffmpeg's AAC implementation, I'm not sure if that applies for decoding or encoding.

I'm pretty sure there are only relevant differences for decoding. Since how the decoded result should look is standardized the only distinguishing features can be features (codec support) or performance (decoding speed).
The other issue here is that mpv will prefer the built-in AAC decoder by default, so users would have to configure `--ad` too for any effect.

Quackdoc:
I'm not sure where to take this then if fdk-aac free doesn't make a lot of sense to include, I could set it up to be locked behind a flag instead of installing fdkaac-free?

sfan5:
Don't think this makes sense then, xHE-AAC seems very niche.

PR #643, opened 2023-01-10
Use LuaJIT instead of Lua52

StratusFearMe21:
I've tested this build on my android phone, and it *does* feel faster (at least to me). One of the benefits of using LuaJIT is that it actually initializes faster than the reference implementation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ addons:
       - libtool
       - ninja-build
       - python3-pip
-      - glibc
+      - libc6-dev
       - python3-setuptools
 
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ addons:
       - libtool
       - ninja-build
       - python3-pip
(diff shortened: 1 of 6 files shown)


StratusFearMe21:
The build commands where adapted from the [luajit website](https://luajit.org/install.html). It's also worth noting that the build I'm using *is* compatible with all the architectures that `mpv-android` supports. The build is a beta build, but Arch Linux uses it, so it's not entirely unstable.

StratusFearMe21:
Hmmm, the error that's happening in Travis CI doesn't happen on my system. It looks like the error is that it's trying to build a binary for the *host* system using the *android* libraries. IDK, maybe someone else knows why it's doing this, CI tools have always been my achilles' heel

sfan5:
Effectively Lua is only useful for two scripts: `auto_profiles.lua` and `stats.lua`, neither of which are in any way performance sensitive.
Due to that and given it doesn't currently build correctly I don't think we need this.

PR #642, opened 2023-01-10
Switch `mpv` build to `meson`

StratusFearMe21:
This seems to be more consistent with the rest of the build scripts, and I found that the build is faster than `waf`. I've tested building with `meson` and it works on my Samsung Galaxy S20 in both `arm64` and `armv7l` modes

PR #644, opened 2023-01-11
Make incremental meson builds faster

StratusFearMe21:
By checking if the build directory already exists, and telling `meson` to `--reconfigure` if it does, it makes multiple successive builds much faster.

### DIFF
--- a/buildscripts/scripts/dav1d.sh
+++ b/buildscripts/scripts/dav1d.sh
@@ -15,8 +15,19 @@ fi
 
 unset CC CXX # meson wants these unset
 
-meson $build --cross-file "$prefix_dir"/crossfile.txt \
-	-Denable_tests=false -Db_lto=true -Dstack_alignment=16
+build_opts=(
+	'-Denable_tests=false'
+	'-Db_lto=true'
+	'-Dstack_alignment=16'
+)
+
+if [ -d $build ]; then
+	build_opts+=("--reconfigure")
+fi
+
+meson setup $build --cross-file "$prefix_dir"/crossfile.txt \
+	"${build_opts[@]}"
+	  
 
 ninja -C $build -j$cores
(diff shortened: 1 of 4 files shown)


sfan5:
I assumed that this is what calling `meson <buildfolder> <options>` did anyway.
If not, does meson have a built-in way to achieve this? Adding this boilerplate in every script is not optimal.

StratusFearMe21:
About the implicit setup command, this is what the meson docs say https://mesonbuild.com/Commands.html#setup
> Deprecated since 0.64.0: This is the default Meson command (invoked if there was no COMMAND supplied). However, supplying the command is necessary to avoid clashes with future added commands, so "setup" should be used explicitly.

StratusFearMe21:
As for the boilerplate, running `meson` with `--reconfigure` on a non-existent or invalid build directory is fatal, and if you run `meson setup` on an already configured build directory, it produces building bugs like it did in my PR, #642. So I'm pretty sure that `meson`  does *not* have a builtin way of doing this.

eli-schwartz:
> and if you run `meson setup` on an already configured build directory, it produces building bugs

Specifically, what it does is print a message stating that it's already configured and therefore will not do anything.

`meson setup --reconfigure builddir -Dopt=val` will force redo the configure process even if the additional option values did not actually change from their previous state, which is potentially slow and a waste of time. `meson configure builddir -Dopt=val` will simply cause `ninja` to first reconfigure as the first step in building. It also has an optimization where it checks if the modified options data has changed, and only updates that if it changed -- so `ninja` sees that nothing changed, and does not reconfigure at all.

StratusFearMe21:
@eli-schwartz, thanks for making me aware of this. I've implemented this in my latest commit.

StratusFearMe21:
Never mind, this seems to have no effect. `meson setup` on an already configured builddir *should* do nothing

PR #645, opened 2023-01-13
Fix meson build issues

StratusFearMe21:
Should fix the static building issue in #642 and make incremental builds faster


sfan5:
This won't update the file if it actually changes. Don't worry I have a commit for this ready myself.

PR #651, opened 2023-01-24
No lag when going into background play on an audio file with album art

StratusFearMe21:
This PR when paired with mpv-player/mpv#11202 will allow the player to enter and exit background play on audio files with no stuttering. For more info, see  mpv-player/mpv#11202


StratusFearMe21:
This PR is not necesary as long as mpv-player/mpv#11202 is applied to the build

StratusFearMe21:
This PR is not necesary as long as mpv-player/mpv#11202 is applied to the build

StratusFearMe21:
Note, this mostly works but it seems that when `thumbnail.cpp` calls `screenshot-raw` and the album art is a `yuv420p` photo, it is expensive enough to pause for more than a second before resuming playback

PR #676, opened 2023-03-12
AndroidManifest: add intent-filter for bilibili

rhjdvsgsgks:


app/src/main/AndroidManifest.xml
```diff
@@ -250,6 +250,24 @@
...
                 <data android:pathPattern=".*" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <data android:scheme="http" />
+                <data android:scheme="https" />
+                <data android:host="bilibili.com" />
+                <data android:pathPattern=".*" />
```
sfan5, line 260: would it make sense to use a pattern like `/video/.*` here?

PR #688, opened 2023-04-17
Add lavf to openurl dialog

langyxxl:
support lavf:// url format.
maybe useful for some advanced use.


langyxxl:
From mpv man, the lavf protocol is used to access different audio and video formats, while the av protocol is used to access libavdevice inputs :
lavf://...
Access any FFmpeg/Libav libavformat protocol. Basically, this passed the string after the // directly to libavformat.

av://type:options avdevice:// is an alias.
This is intended for using libavdevice inputs. type is the libavdevice demuxer name, and options is the filename passed to the demuxer.

sfan5:
I see. However you will also need to change `MPVActivity.resolveUri()` for this to work.

PR #715, opened 2023-05-18
Make watch-later directory user accessible

syphyr:
This fixes resuming playback when watch-later is enabled by creating a user accessible directory on external storage (/sdcard/watch_later). This also allows the user to easily cleanup orphaned resume files.

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVView.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVView.kt
@@ -9,6 +9,7 @@ import android.os.Build
 import android.os.Environment
 import android.preference.PreferenceManager
 import android.view.*
+import java.io.File
 import kotlin.math.abs
 import kotlin.reflect.KProperty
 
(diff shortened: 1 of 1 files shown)


qgustavor:
Was it tested on Android 11 (or above) or just Nougat? I might be wrong, but looking at the diff it makes me think it solves the issue on older Android but makes it worse on newer Android.

syphyr:
> Was it tested on Android 11 (or above) or just Nougat? I might be wrong, but looking at the diff it makes me think it solves the issue on older Android but makes it worse on newer Android.

Tested on Nougat, but using the build for API 31.

syphyr:
The screenshot directory is also created in the same way as the watch-later directory, so I would assume all permissions are the same as before.

syphyr:
changed directory from /sdcard/mpv-android to /sdcard/watch_later

biopsin:
How about exsisting  /sdcard/Android/media/is.xyz.mpv ?
not big fan of uninportant folder/file cluttering root of sdcard

lntls:
Why not ```Context.getCacheDir```?

syphyr:
> How about exsisting /sdcard/Android/media/is.xyz.mpv ? not big fan of uninportant folder/file cluttering root of sdcard

I like this suggestion best.  I have changed it to /sdcard/Android/media/is.xyz.mpv/watch_later.

syphyr:
Used context.getPackageName() instead of hardcoded value.

sfan5:
This is now fixed inside libmpv, no need for any changes.

PR #743, opened 2023-07-07
enable ipfs protocols (ipfs and ipns)

markg85:
Allows to type a URL like:
ipfs://<cid>
ipns://<cid>

mpv already allows this, just mpv-android didn't.

I'm intentionally keeping this commit clean and simple even though this commit on it's own isn't that useful. A user still can't actually do `ipfs://<cid>` because a gateway is needed for that. I'll make a separate PR for that. It's all already possible by modifying mpv.conf though (that pr is just going to make it more easier to set).

### DIFF
--- a/app/src/main/java/is/xyz/mpv/Utils.kt
+++ b/app/src/main/java/is/xyz/mpv/Utils.kt
@@ -412,6 +412,6 @@ object Utils {
     // cf. AndroidManifest.xml and MPVActivity.resolveUri()
     val PROTOCOLS = setOf(
         "file", "content", "http", "https",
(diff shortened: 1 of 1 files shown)


markg85:
ping @sfan5

sfan5:
Sorry for ignoring this until now. I am personally skeptical of the real world usability/practically of IPFS, so I won't be merging this.

markg85:
Does it make sense in trying to convince you or is your mind set on a hard no-go?

sfan5:
More of the latter.

markg85:
Annoying but fair enough ;)

PR #773, opened 2023-09-03
Add Polish translation

Matthaiks:


sfan5:
merged as 6c238bc, thanks

PR #849, opened 2024-01-12
Add force hardware plus detection setting

syphyr:
Mpv is unable to automatically detect hardware plus decoding on my device without manually editing mpv's config.  This PR adds a checkbox underneath the current Hardware Decoding setting to forcibly enable hardware plus decoding. `hwdec` is set to "mediacodec" and `vo` is set to "mediacodec_embed".


sfan5:
But `mediacodec_embed` isn't "force hardware decoding", it's a totally different VO with notable features missing and AFAIR it won't even fallback to swdec correctly.
This is not something I want to offer to users.

syphyr:
I understand. I'll close it.

PR #903, opened 2024-04-24
Make path.sh to dynamically use the NDK version specified in depin…

gakramx:
Remove hardcoded NDK version and replaced it with a reference to the version defined in depinfo.sh

### DIFF
--- a/buildscripts/include/path.sh
+++ b/buildscripts/include/path.sh
@@ -2,6 +2,8 @@
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && cd .. && pwd )"
 
+. $DIR/include/depinfo.sh
+
 os=linux
 [[ "$OSTYPE" == "darwin"* ]] && os=mac
 export os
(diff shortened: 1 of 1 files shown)


buildscripts/include/path.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
 
+# Get the directory of the current script
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+. "$SCRIPT_DIR/depinfo.sh"
```
sfan5, line 6: this is a bit redundant, you can just use `$DIR/include/depinfo.sh`

gakramx: Thanks for the review.
I opted for `$SCRIPT_DIR/depinfo.sh` over `$DIR/include/depinfo.sh `to ensure the script's portability and reliability.

 Using `$SCRIPT_DIR` guarantees that the script works regardless of its location, enhancing readability by explicitly stating the dependency's location relative to the script. 

Would you like me to continue with this approach, or do you have a preference for using `$DIR`?

sfan5: It's not like this script is designed to be moved, so yes I prefer relying on `$DIR`.

gakramx: Thank you i get it , I will update it now.

sfan5:
Looks good, can you squash the commits?

gakramx:
> Looks good, can you squash the commits?

Sorry i forget this , I squashed 3 commits into one !

syphyr:
I'm not sure why $DIR is needed. All the other scripts just use `. ./include/depinfo.sh`

sfan5:
merged as 1b8d2935229f607c12e7b26863a741bbec6d5a99

gakramx:
> I'm not sure why $DIR is needed. All the other scripts just use `. ./include/depinfo.sh`

Using `$DIR` ensures consistent and reliable sourcing of **depinfo.sh**, regardless of the script's location.

PR #935, opened 2024-06-02
MPVActivity: prevent screen timeout

syphyr:
On MPV_EVENT_PLAYBACK_RESTART, it is necessary in certain cases to call updatePlaybackStatus so that FLAG_KEEP_SCREEN_ON is added, otherwise this windowmanager flag may never get set and the screen times out during playback.

This reverts commit 6902b73fb5be86e4306bda2dabc58d89451f1bd3.

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
@@ -1699,7 +1699,9 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
 
     private fun eventUi(eventId: Int) {
         if (!activityIsForeground) return
-        // empty
+        when (eventId) {
+            MPVLib.mpvEventId.MPV_EVENT_PLAYBACK_RESTART -> updatePlaybackStatus(player.paused!!)
+        }
     }
 
     override fun eventProperty(property: String) {


sfan5:
This smells like the wrong fix.
What are these "certain cases"?
What's the root cause? Is `eventPropertyUi` not being called? Or are the updates somehow missed and not re-done (e.g. activity backgrounded)?

syphyr:
After 6902b73fb5be86e4306bda2dabc58d89451f1bd3 was applied, I noticed that when I initially played a video from a file explorer, the screen timed out during playback after a few minutes.  After I woke the device up by pressing the home button, then the video would play to the end of the file and no longer timeout during playback.  It only seemed to happen on the initial start of playback. Although, sometimes I found it difficult to reproduce.  Nonetheless, after reverting that change mentioned above, it no longer happens.

sfan5:
It would be nice if you could re-test with these changes:
```diff
diff --git a/app/src/main/java/is/xyz/mpv/MPVActivity.kt b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
index bc0dfff..9f7420d 100644
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
@@ -1418,6 +1418,7 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
 
     private fun refreshUi() {
         // forces update of entire UI, used when resuming the activity
+        Log.v(TAG, "updatePlaybackStatus(${psc.pause}) via refreshUi")
         updatePlaybackStatus(psc.pause)
         updatePlaybackPos(psc.positionSec)
         updatePlaybackDuration(psc.durationSec)
@@ -1675,7 +1676,10 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
     private fun eventPropertyUi(property: String, value: Boolean) {
         if (!activityIsForeground) return
         when (property) {
-            "pause" -> updatePlaybackStatus(value)
+            "pause" ->  {
+                Log.v(TAG, "updatePlaybackStatus($value) via eventPropertyUi")
+                updatePlaybackStatus(value)
+            }
         }
     }
 
@@ -1716,6 +1720,8 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
     }
 
     override fun eventProperty(property: String, value: Boolean) {
+        if (property == "pause")
+            Log.v(TAG, "eventProperty($property, $value)")
         if (psc.update(property, value))
             updateMediaSession()
         if (property == "shuffle") {
```

sfan5:
Nevermind, I stumbled upon the bug you mean today myself.

syphyr:
I'm glad you are able to reproduce as well.  So, `FLAG_KEEP_SCREEN_ON` was previosly set by `eventPropertyUi`, `eventUi`, and `refreshUi` when but bug did not exist.  But, `refreshUi` is only called when resuming an activity, and this bug only seems to happen when a file is initially played, not when it is resumed.  So, there must be some condition that `eventUi` is called when `eventPropertyUi` is not on initial playback.  The main problem is that I have a difficult time reproducing this.

PR #985, opened 2024-09-03
install meson from debian repository

K-os:
Because Debian no longer allows installing packages system-wide using pip.


buildscripts/include/download-sdk.sh
```diff
@@ -15,8 +15,7 @@ if [ "$os" == "linux" ]; then
 			pip3 install -U meson
 		elif apt-get -v &>/dev/null; then
 			sudo apt-get install autoconf pkg-config libtool ninja-build \
-			python3-pip unzip wget
-			pip3 install -U meson
+				python3-pip unzip wget meson
```
sfan5, line 18: can remove pip then

PR #990, opened 2024-09-07
Small fixes for English, Ukrainian and Russian translations

bohdangnatiuk:
"Panscan" is the name of the "Pan and Scan" function. And it's called exactly "Panscan". You can confirm it by turning it on or off. Therefore, it's best not to invent or rename it.

As for the Ukrainian or Russian translation, these are the languages I'm proficient in. It's quite pointless to translate this function directly, as it's similar to translating the word Wi-Fi. In summary — it isn't worth translating.

### DIFF
--- a/app/src/main/res/values-ru/arrays.xml
+++ b/app/src/main/res/values-ru/arrays.xml
@@ -6,7 +6,7 @@
         <item>16:10</item>
         <item>4:3</item>
         <item>2.35:1</item>
-        <item>Panscan</item> <!-- TODO -->
+        <item>Panscan</item>
     </string-array>
     <string-array name="background_play_entries">
         <item>Никогда</item>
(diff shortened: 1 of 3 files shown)


sfan5:
The mpv manual consistently calls it "pan-and-scan" but whatever, this should be fine.

app/src/main/res/values-ru/arrays.xml
```diff
@@ -6,7 +6,7 @@
         <item>16:10</item>
         <item>4:3</item>
         <item>2.35:1</item>
-        <item>Pan/Scan</item> <!-- TODO -->
+        <item>Panscan</item> <!-- TODO -->
```
sfan5, line 9:
```suggestion
        <item>Panscan</item>
```

bohdangnatiuk:
> The mpv manual consistently calls it "pan-and-scan" but whatever, this should be fine.

Yeah, I saw. But for some reason, they call it "Panscan" in the player itself when you turn it on. "Pan and Scan" is more likely used to describe the function itself. To me, "Panscan" sounds simpler and easier. The current version with the split through "/" looks a bit odd. So it would be better to use "Pan and Scan", or even better "Panscan", since that’s what it's called in the app itself.

PR #992, opened 2024-09-14
feat: Add themed/monochrome icon 

### LINKED ISSUE #973: Material You Icon
It would be great if the mpv icon could be themed with Material You, as it would help it to blend in with the rest of my setup (and, no doubt, other people's setups also).

khaled-0:
image

Slight variation of: https://github.com/LawnchairLauncher/lawnicons/blob/develop/svgs/mpv_android.svg
 (close #973 )

### DIFF
--- a/app/src/main/res/drawable/mpv_monochrome.xml
+++ b/app/src/main/res/drawable/mpv_monochrome.xml
@@ -1,0 +1,3 @@
+<inset xmlns:android="http://schemas.android.com/apk/res/android"
+    android:drawable="@drawable/ic_mpv_symbolic"
+    android:inset="26dp" />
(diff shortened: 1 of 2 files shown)


sfan5:
How is that file licensed? Same as mpv-symbolic.svg since it's a derivative I assume.

khaled-0:
> How is that file licensed? Same as mpv-symbolic.svg since it's a derivative I assume.
It's Apache 2.0

holup, I didn't see that file.
Does it look better or worse? Better imo
image

sfan5:
I would prefer mpv-symbolic since it's an official part of mpv.

khaled-0:
> I would prefer mpv-symbolic since it's an official part of mpv.

agree, done

sfan5:
Thanks.

PR #994, opened 2024-09-16
buildscripts: Don't clone dependency history

khaled-0:
Reduced storage and network usage
We don't need git histories
https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/


obscenelyvague:
Would make working on the dependencies alone needlessly difficult. For example, if there were a regression and you had to bisect.

khaled-0:
> Would make working on the dependencies alone needlessly difficult. For example, if there were a regression and you had to bisect.

Then how about `` --filter=blob:none `` 
Also why tf ic CI failing, I'm sure it has nothing to do with this commit

khaled-0:
also how about implementing dependency caching

sfan5:
> Also why tf ic CI failing, I'm sure it has nothing to do with this commit ?? or is it!!

yes it does. `--depth=1` breaks checking out ffmpeg

> also how about implementing dependency caching

The code tries to intelligently cache dependencies so they're not rebuilt every time, but this is currently broken for some reason.

khaled-0:
> > Also why tf ic CI failing, I'm sure it has nothing to do with this commit ?? or is it!!
> 
> yes it does. `--depth=1` breaks checking out ffmpeg


How about ``--filter=blob:none`` 
ci seems to pass, 
It'll skip downloading old blobs from git history

PR #1049, opened 2024-12-20
Update Russian translation

solokot:


app/src/main/res/values-ru/strings.xml
```diff
@@ -4,34 +4,32 @@
...
     <string name="dialog_yes">Да</string>
     <string name="dialog_no">Нет</string>
     <string name="dialog_prev">Назад</string>
-    <string name="dialog_next">Вперед</string>
+    <string name="dialog_next">Вперёд</string>
+    <string name="dialog_reset">Сброс</string>
     <string name="track_audio">Аудио</string>
     <string name="track_subs">Субтитры</string>
-    <string name="track_off">Выкл</string>
+    <string name="track_off">ОТКЛ</string>
```
sfan5, line 18: going by a dictionary the old form should be correct, it's supposed to mean "off, disabled, inactive"

sfan5 (commented):
looks good otherwise

varto150 (approved):

app/src/main/res/values-ru/strings.xml
solokot: Hello!
"откл" & "выкл" are abbreviations for synonymous words, both meaning "disabled". It is preferable to use "(откл)ючено" so that the word is visually different from "(вкл)ючено" meaning "enabled", "on".
But specifically in this case, if the "Off" string is not used except as an indication of turning off the track, it is better to change it to "откл.", i.e.:
<string name="track_off">откл.</string>

PR #1062, opened 2025-01-27
fixed the problem gradle not building on new version on android studio

lotfi-workbox:


lotfi-workbox:
there is problem on base origin with gradle that can be fix with adding that two line

app/build.gradle
```diff
@@ -47,6 +47,13 @@ android {
         }
     }
 
+    sourceSets {
+        main {
+            jniLibs.srcDirs = ['../../sdk/native/libs/']
```
sfan5, line 52: this doesn't look right to me, there's no such folder.
you can see that it breaks the CI build too

lotfi-workbox: do you use any jniLibs in application?

lotfi-workbox: i checked your ci build dependencies it seems to your sdk path is /usr/local/lib/android/sdk so i changed it

fangfengrui: 3q，it works right

PR #1066, opened 2025-02-14
MPVActivity: fix audio focus not being requested on resume (closes #1063)

### LINKED ISSUE #1063: Take audio focus based on actual playback
Whether to take audio focus is decided only when starting player. This makes it so that audio focus can no longer be attained when player is already open. ~~Additionally, maybe it should always be ignored if an audio track isn't present or at least for images specifically.~~ This part is not as important.

Tigermouthbear:
Before this patch audio focus was only gained when playback first started. If you paused then resumed playback audio focus would not be gained, causing other apps to continue playing audio in the background. This patch fixes this behavior while following the Ignore Audio Focus setting.

closes #1063

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
@@ -336,24 +336,24 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
 
         volumeControlStream = STREAM_TYPE
 
-        // Handle audio focus
-        val req = with (AudioFocusRequestCompat.Builder(AudioManagerCompat.AUDIOFOCUS_GAIN)) {
-            setAudioAttributes(with (AudioAttributesCompat.Builder()) {
-                // N.B.: libmpv may use different values in ao_audiotrack, but here we always pretend to be music.
-                setUsage(AudioAttributesCompat.USAGE_MEDIA)
-                setContentType(AudioAttributesCompat.CONTENT_TYPE_MUSIC)
+        // Handle audio focus, see also eventProperty for audio focus handling on pause and resume and onDestroy for cleanup
+        if (!ignoreAudioFocus) {
+            audioFocusRequest = with (AudioFocusRequestCompat.Builder(AudioManagerCompat.AUDIOFOCUS_GAIN)) {
+                setAudioAttributes(with (AudioAttributesCompat.Builder()) {
+                    // N.B.: libmpv may use different values in ao_audiotrack, but here we always pretend to be music.
+                    setUsage(AudioAttributesCompat.USAGE_MEDIA)
+                    setContentType(AudioAttributesCompat.CONTENT_TYPE_MUSIC)
+                    build()
+                })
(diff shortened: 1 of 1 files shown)


sfan5:
> If you paused then resumed playback audio focus would not be gained, causing other apps to continue playing audio in the background.

Is the audio focus permanently lost when pausing or why does this happen? Or in other words, does `AUDIOFOCUS_LOSS` mean we have to request it all over again?
As far as I understand as long as the app gains focus at startup it will keep a "handle" on it so that `OnAudioFocusChangeListener` can handle any events that my might occur.
The [docs](https://developer.android.com/media/optimize/audio-focus) seem to agree:

"Your app doesn't have to abandon audio focus if the user pauses playback but might resume playback later."

Tigermouthbear:
OnAudioFocusChangeListener doesn't actually do anything in Android 12+, this is all managed by the system now. However, you always had to request audio focus if your app does not currently have it. Audio focus is lost when you start playing audio from another app (like playing music or a YouTube video). If this happens and you resume playback without requesting audio focus again, then the other app will keep audio focus and mpv will play over the app with audio focus. Therefore, to prevent this you can just request audio focus every time playback resumes.

As for abandoning audio focus on pause, the docs make it seem like before Android 12 your app should release audio focus but its not very clear on when. It might not necessary on pause, but the docs also say that this method "Causes the previous focus owner, if any, to receive focus." So I figured that it might be beneficial to add it.

sfan5:
Okay looks like the app should indeed re-request the focus for every operation that begins playing.

> OnAudioFocusChangeListener doesn't actually do anything in Android 12+, this is all managed by the system now. 

I'm aware that the OS just forces the volume change onto you, but surely the callback will still be used to tell the app that it has lost audio focus, won't it?

app/src/main/java/is/xyz/mpv/MPVActivity.kt
```diff
@@ -1784,13 +1784,19 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
...
+
         if (property == "shuffle") {
             mediaSession?.setShuffleMode(if (value)
                 PlaybackStateCompat.SHUFFLE_MODE_ALL
             else
                 PlaybackStateCompat.SHUFFLE_MODE_NONE)
         }
 
+        if (property == "pause" && !ignoreAudioFocus && audioFocusRequest != null) {
+            if (!value) AudioManagerCompat.requestAudioFocus(audioManager!!, audioFocusRequest!!)
```
sfan5, line 1796: what should happen here if the audio focus is not granted?

sfan5:
> [...] but surely the callback will still be used to tell the app that it has lost audio focus, won't it?

By experimentation this seems to be true at least on Android 14.

Tigermouthbear:
Can confirm your commit is working as intended! Sorry for not responding to your review quickly.

PS: The reason the requesting doesn't work in eventProperty at the beginning is because eventProperty("pause", false) is called before MPVActivity::onCreate, so the audio manager is null when this happens.

PR #1073, opened 2025-02-25
Fix clone directory name

syphyr:
The script checks for "freetype2" directory, so fix the name when cloning freetype to match previous check.


sfan5:
oops. merged as d7a4e75f706bb9053faf56f0bb15a3eac76d6694, thanks.

PR #1087, opened 2025-04-14
MPVActivity: Close file descriptors when no longer needed

chenxiaolong:
3ada286e34625068bd74370db7cf413614e5c7df adjusted things so that `fd://` URIs are used instead of `fdclose://` URIs due to the latter not being reusable when using playlists. However, this left no mechanism to close file descriptors that are no longer needed, which can cause SAF content providers to keep running in the background indefinitely.

This commit updates `MPVActivity`'s logic to keep track of open file descriptors and close them when they are no longer needed. This is primarily when `MPVActivity` is destroyed and when handling a new intent that does not append to the playlist.

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
@@ -46,6 +46,7 @@ import androidx.media.AudioAttributesCompat
 import androidx.media.AudioFocusRequestCompat
 import androidx.media.AudioManagerCompat
 import java.io.File
+import java.io.IOException
 import java.lang.IllegalArgumentException
 import kotlin.math.roundToInt
 
@@ -77,6 +78,8 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
 
     private val psc = Utils.PlaybackStateCache()
     private var mediaSession: MediaSessionCompat? = null
+
+    private val openPfds = mutableSetOf<ParcelFileDescriptor>()
 
     private lateinit var binding: PlayerBinding
     private lateinit var gestures: TouchGestures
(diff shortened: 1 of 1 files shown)


app/src/main/java/is/xyz/mpv/MPVActivity.kt
```diff
@@ -363,6 +371,9 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
             moveTaskToBack(true)
         } else {
             MPVLib.command(arrayOf("loadfile", filepath))
+            // When not appending to a playlist, there's no reason to keep old files open anymore
+            previousPfds.forEach { it.close() }
+            openPfds.removeAll(previousPfds)
```
sfan5, line 376: mpv commands are async so it might still be using the fd when this runs.
only safe thing to do it keep them...

chenxiaolong: Ah shoot, I didn't realize that.

chenxiaolong: Oh, a custom protocol handler sounds like a great idea. I didn't realize mpv could do this without changing its core code, but it looks like `mpv/stream_cb.h` is exactly for this purpose. I'll try implementing this approach this weekend.

sfan5:
07726e6840099ff8f65a1140a82c4df61974c76f. thanks for contributing.

app/src/main/java/is/xyz/mpv/MPVActivity.kt
chenxiaolong: I've opened https://github.com/mpv-android/mpv-android/pull/1090. It turns out ffmpeg 7 natively supports `content://` URIs as long as it has access to the application context, so there was no need to implement a custom protocol handler.

PR #1102, opened 2025-04-30
KeyMapping: add mapping for colour keys

TheDcoder:
The keys are mapped to the function keys which are analogous to the coloured programmable function keys in a TV remote.

Side-note: Is there a good place to document this for the end user? Perhaps we can add an input test mode in the app to let the user find out the key mappings via experimentation.


sfan5:
> Is there a good place to document this for the end user?

Not sure about good but we could add a page to the [docs](https://github.com/mpv-android/mpv-android/tree/master/docs) folder and have the app link to there.

TheDcoder:
That sounds good enough to me but the problem is I don't know how to do the last part, I am not an Android developer :sweat:

PR #1113, opened 2025-07-05
Turkish Language

fuckpdf:
Some minor fixes


app/src/main/res/values-tr/arrays.xml
```diff
@@ -1,33 +1,120 @@
...
         <item>CPU</item>
         <item>GPU</item>
     </string-array>
+    <string-array name="deband_values" translatable="false">
+        <item></item>
+        <item>gradfun</item>
+        <item>gpu</item>
+    </string-array>
+
+    <string-array name="video_sync" translatable="false">
```
sfan5, line 108: entries with `translatable="false"` do not need to be copied

PR #1116, opened 2025-07-13
add persian translation

eshagh79:
Hello, I translated mpv into Persian, please add it to the program.
I also translated the name of the program, but this translation gives the exact same sound as m. p. v. If this is not what you want, I can return it to the same English form.


app/src/main/res/values-fa/arrays.xml
```diff
@@ -0,0 +1,159 @@
...
+    <!-- Arrays used by the UI -->
+    <string-array name="aspect_ratio_names">
+        <item>اصلی</item>
+        <item>Panscan</item>
+        <item>۱۶:۹</item>
+        <item>۱۶:۱۰</item>
+        <item>۴:۳</item>
+        <item>۲.۳۵:۱</item>
+    </string-array>
+    <string-array name="aspect_ratios" translatable="false">
```
sfan5, line 12: please remove the entries with `translatable="false"`, no need to duplicate them.

sfan5:
> I also translated the name of the program, but this translation gives the exact same sound as m. p. v. If this is not what you want, I can return it to the same English form.

Not sure. What do other apps usually do? Keep the Latin letters?

eshagh79:
Translating an app name is more a matter of the developer's decision. For example, Firefox has prohibited this in its policy, or Telegram has only made the app name untranslatable but used Telegram's translation within its app, while F-Droid and most of the apps within it have allowed app name translation.
VLC has also made the program name untranslatable, but still allows translation within the program.
In my opinion, there is no problem with translating specific names like MPV literally.

sfan5 (approved):
Tested this a bit and I think the LTR layout isn't correct in all places the translated parts do look correct.

sfan5:
b37f1564de1832efec43962787855c09b077c714. thanks.

eshagh79:
> Tested this a bit and I think the LTR layout isn't correct in all places the translated parts do look correct.

Hi, I can't confirm whether all Persian language strings are displayed correctly or not. Because I don't have the beta version of this program and building it from source code is too complicated for me.

Anyway, I added Persian language translation to Android MPV with the apkeditor program, the strings are displayed correctly, but they are all aligned to the left, which is wrong.

If it's not too much trouble for you, create a beta version for me to review before the final release.
If you also create a new issue for bugs in right-to-left languages (such as Persian, Arabic, Urdu, Hebrew, etc.) we can continue the conversation there.

sfan5:
You can download an APK from here: https://github.com/mpv-android/mpv-android/actions/runs/16367988229

PR #1118, opened 2025-07-22
Add support for right-to-left languages

eshagh79:
Hello.
In this request, I enabled the right-to-left language support feature so that when the program starts with languages such as Persian, the settings and text are aligned on the right.
When this feature is enabled, when displaying a video, the progress bar moves from right to left, which does not match the hand gestures, and the control buttons are also aligned on the right, which also causes the previous and next track buttons to be displayed incorrectly when playing a music list.


sfan5 (approved):
didn't test `fa-IR` but no change for the english UI

sfan5:
bc4b9a37755a758eae9ca3ed7ed9d2a09dd8f6d1

PR #1119, opened 2025-07-24
Buildscripts: add a docker buildscript

abdallahmehiz:
Implements a few scripts to build mpv-android inside a docker container.
running `docker-build.sh` would create the docker image, mount the mpv-android project from the host system then starts the build process from `docker-entrypoint.sh`.


kraveyard:
@sfan5, any update on this?

sfan5:
While I can see mpv-android shipping with a helper to make setting up docker easier I don't see why it's necessary to essentially duplicate half of the build-related scripts to enable this.

> @sfan5, any update on this?

Update on what? If you want to use these scripts you can do that today. No need to wait for me.

abdallahmehiz:
> While I can see mpv-android shipping with a helper to make setting up docker easier I don't see why it's necessary to essentially duplicate half of the build-related scripts to enable this.

I've rewritten it so all it does is forward download or build commands to `buildall.sh` or `download.sh`.
The only additional thing is the `shell` argument to launch a shell inside the docker container with everything mounted.

kraveyard:
> Update on what? If you want to use these scripts you can do that today. No need to wait for me.

Just wanted to know if there is a plan to merge this.

buildscripts/Dockerfile
```diff
@@ -0,0 +1,55 @@
+FROM ubuntu:24.04
+
+# Set up environment variables
+ENV DEBIAN_FRONTEND=noninteractive
+ENV ANDROID_HOME=/opt/android-sdk
```
sfan5, line 5: is the user expected to have the sdk installed system-wide? is this a standard path?

abdallahmehiz: No, the standard path is `~/Android/Sdk`.
That's where the Android SDK defaults to.

abdallahmehiz: It's a container, so /opt made more sense to me than a user home path.

sfan5:
I don't think this is useful enough, sorry.

PR #1120, opened 2025-07-24
MPVLib: null-safe property getters

abdallahmehiz:


abdallahmehiz:
I am also planning on PRing https://github.com/abdallahmehiz/mpv-android/commit/fe8f4555fcfe9ed125f7038367f8073265fc8b04, but I'll wait for a response to this PR before I do that

app/src/main/java/is/xyz/mpv/MPVLib.kt
```diff
@@ -0,0 +1,176 @@
...
+import android.content.Context
+import android.graphics.Bitmap
+import android.view.Surface
+
+// Wrapper for native library
+
+@Suppress("unused")
+object MPVLib {
+    init {
+        val libs = arrayOf<String?>("mpv", "player")
```
sfan5, line 12:
```suggestion
        val libs = arrayOf("mpv", "player")
```

app/src/main/java/is/xyz/mpv/MPVLib.kt
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    external fun create(appctx: Context)
+    external fun init()
+    external fun destroy()
+    external fun attachSurface(surface: Surface)
+    external fun detachSurface()
+
+    external fun command(cmd: Array<String>)
```
sfan5, line 24: I think this is correct?
```suggestion
    external fun command(cmd: Array<out String>)
```

app/src/main/java/is/xyz/mpv/MPVLib.kt
```diff
@@ -0,0 +1,176 @@
...
+    external fun getPropertyDouble(property: String): Double?
+    external fun setPropertyDouble(property: String, value: Double)
+    external fun getPropertyBoolean(property: String): Boolean?
+    external fun setPropertyBoolean(property: String, value: Boolean)
+    external fun getPropertyString(property: String): String?
+    external fun setPropertyString(property: String, value: String)
+
+    external fun observeProperty(property: String, format: Int)
+
+    private val observers: MutableList<EventObserver> = ArrayList<EventObserver>()
```
sfan5, line 41:
```suggestion
    private val observers = mutableListOf<EventObserver>()
```

app/src/main/java/is/xyz/mpv/MPVLib.kt
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    @JvmStatic
+    fun event(eventId: Int) {
+        synchronized(observers) {
+            for (o in observers) o.event(eventId)
+        }
+    }
+
+    private val log_observers: MutableList<LogObserver> = ArrayList<LogObserver>()
```
sfan5, line 95: same here

sfan5 (commented):
question: would `MPVLib` remain usable from Java like this?

app/src/main/java/is/xyz/mpv/MPVLib.kt
abdallahmehiz: Both would work, but if you want to accept subtypes of `String` like `Nothing`, which... I don't really find a use for in this case, then yes, the `Array<out String>` would also be correct

abdallahmehiz:
> question: would `MPVLib` remain usable from Java like this?

Yes, It'd work as if it's still a .java file

sfan5:
merged as 0a68cb376fbdb673ca0152c72dd0d9766e077563, thanks.

PR #1149, opened 2025-10-15
MPVActivity: don't switch to the audio ui when a video file/stream is loading

puchu:
MPVActivity: don't switch to the audio ui when a video file/stream is loading


sfan5:
I'd like to solve this a bit differently, but no idea how yet.

puchu:
how about applying the pr until you find a better solution?

sfan5:
In its current state there is a risk that the audio UI state is not correctly updated afterwards because `vid`/`aid` changing won't re-trigger the code.

Issue #151, opened 2018-11-01
Open an audio track in the background by default


ArenaL5:
First of all, thank you all very much for porting mpv to android. mpv is currently my media player of choice in Windows, Linux and Android.

This question is related to both [issue #102](https://github.com/mpv-android/mpv-android/issues/102) and [#104](https://github.com/mpv-android/mpv-android/issues/104).

Right now, I've enabled the option to continue playback in the background, but when I open an audio track (MP3 or whatever), MPV opens a black screen like in the PC. I need to press Home or Switch App to resume what I was doing before.

I'd like to ask for a couple of small improvements that would improve a lot the Android experience:
- upon opening a file with one audio track and no video tracks, automatically open mpv in the background, and allow it to be controlled via the notification as usual. (Or, if people don't consider this a good default, allow some way to enable in the configuration, preferably via the GUI.)
- a play/pause button in the notification, if possible.

I also have a question: what are the requirements for the playlist buttons appear?

I'm using an Android Marshmallow mobile phone, if it helps.

sfan5:
> I also have a question: what are the requirements for the playlist buttons appear?

Playlist buttons appear whenever you open something mpv considers a playlist (and has more than 1 entry).
This is usually achieved by opening an .m3u file or opening a folder as playlist by long-pressing it.

PR #1158, opened 2025-10-27
background patch for audio

piyushgarg:
fixing [151](https://github.com/mpv-android/mpv-android/issues/151)


piyushgarg:
@sfan5 It is still trying to open something in background for a flash and then goes background. How is it possible to run in complete background?

sfan5:
The activity will somehow need to receive the intent without being brought to front by Android, I don't know how to make that work and haven't investigated this feature request.

Issue #979, opened 2024-08-21
Missing DASH support


lodgerz:
hello, there is a streaming tv that uses .mpd files. On mpv for pc it works but on android it can't open it.
Example:
https://live03-col.msf.cdn.mediaset.net/live/ch-ki/ki-clr.isml/manifest.mpd

sfan5:
mpv-android is not built with libxml, which ffmpeg uses for DASH support.

okcaptain:
> mpv-android is not built with libxml, which ffmpeg uses for DASH support.

It would be great to be able to play mpd files

Issue #1169, opened 2025-11-14
Malayalam subtitle is not rendering properly (missing font support)


sanelysan:
Important Information

Android Version: Android 15
mpv-android Version: 2025-08-25-release-oldapi / 8043 (release)
mpv v0.40.0 built on Aug 25 2025
Description

When I load a video with Malayalam subtitles (either external .srt or embedded), the text does not render. The subtitle area is either completely blank or shows boxes/unrendered characters.
This seems to be arendering issue with the Malayalam script (മലയാളം).

Steps to [Reproduce:]

Open mpv-android.
Load a video with a selected Malayalam subtitle track.
Subtitles fail to appear or are corrupted.
Here's the malayalam.srt
https://drive.google.com/file/d/1_LDCT6qUjhOppyWF_WXqP6cHW6aOkNv7/view?usp=drivesdk

Additional Information

Here's the screenshot 

![malayalam subtitle rendering issue](https://github.com/user-attachments/assets/22ac8dd1-d912-4152-834a-9b0f05eb6659)

Sensex00:
This is not a problem with mpv rather the reason is default font that mpv uses subfont.ttf doesn't supports malaylam  
So I think you have find and use a font which supports Malayalam like noto sans malyalam https://fonts.google.com/noto/specimen/Noto+Sans+Malayalam
You should download it , extract,go into the static and get the noto sans malyalam bold.
Then Follow steps here https://github.com/mpv-android/mpv-android/issues/1048#issuecomment-3340564220
Make a folder named fonts in android/media/is.xyz.mpv/fonts

Then Copy that font file into the fonts folder 
And simply paste these into mpv.conf

sub-fonts-dir=/storage/emulated/0/android/media/is.xyz.mpv/fonts

sub-ass-override=force
sub-font="Noto Sans Malayalam Bold"
sub-font-size=53
sub-color="#FFFFFF"
sub-margin-y=40
sub-border-size=2.4
sub-border-color="#FF000000"
sub-shadow-color="#A0000000"
sub-shadow-offset=0.75
sub-ass-style-overrides=Kerning=yes

Issue #1072, opened 2025-02-24
Bangla Subtitle Issue


sheikhimoon:
Is there any way to add Bangla font for subtitle? 
Facing □□□□□ this issue.

sheikhimoon:
> Need sample file.

https://files.catbox.moe/pffbyp.srt

sheikhimoon:
> Need sample file.

Any solution?

sheikhimoon:
> [@sheikhimoon](https://github.com/sheikhimoon) create a folder "fonts" in your mpv directory (android/media/is.xyz.mpv) and place there Bangla.ttf and then in mpv.conf add this lines:
> 
> sub-fonts-dir=/storage/emulated/0/android/media/is.xyz.mpv/fonts sub-font="Bangla"
> 
> You can correct path according to your's device

@dexaltat Not working. 

![Image](https://github.com/user-attachments/assets/faf43c3d-09e7-4ad4-a7b3-2f97bb691763)

![Image](https://github.com/user-attachments/assets/2137dc11-8448-46e3-9513-7bea6d296cce)

dexaltat:
idk, try different ttf file. It's working for me![Screenshot_20250225_044632_mpv.jpg](https://github.com/user-attachments/assets/f095c2db-664c-4408-ac56-1094506e3f3f)

sheikhimoon:
> idk, try different ttf file. It's working for me![Screenshot_20250225_044632_mpv.jpg](https://github.com/user-attachments/assets/f095c2db-664c-4408-ac56-1094506e3f3f)
> 
> 
> 
> 

Which font it is?

sheikhimoon:
> idk, try different ttf file. It's working for me![Screenshot_20250225_044632_mpv.jpg](https://github.com/user-attachments/assets/f095c2db-664c-4408-ac56-1094506e3f3f)
> 
> 
> 
> 

@dexaltat tried different font not working. and didn’t see any .conf file in data and media folder. 
Also trying on another phone. Same issue. 

[Sorry for my English]

sheikhimoon:
@dexaltat @sfan5 please response. Tried SolaimanLipi, SiyamRupali, Kalpuruph font still not working. Even I tried to change different English font but not worked. I think it's directory issue. @dexaltat which font did you use? And can you show me the config text?

sfan5:
[...]

PR #1218, opened 2026-02-25
Build libass with fontconfig support

moi15moi:
libass will use fontconfig to find fonts installed into the system. Without this, libass couldn't find system fonts, so it was always using the "subfont.ttf" setted by mpv by default.

Fix:
- https://github.com/mpv-android/mpv-android/issues/1169
- https://github.com/mpv-android/mpv-android/issues/1072

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install deps
         run: |
           sudo apt-get update
-          sudo apt-get install autoconf pkg-config libtool ninja-build python3-pip
+          sudo apt-get install autoconf pkg-config libtool ninja-build python3-pip gperf
           sudo pip3 install meson
 
       - name: Download deps
--- a/buildscripts/include/depinfo.sh
+++ b/buildscripts/include/depinfo.sh
@@ -15,6 +15,8 @@ v_harfbuzz=12.3.2
 v_fribidi=1.0.16
 v_freetype=2.14.1
 v_mbedtls=3.6.5
+v_libxml2=2.15.2
+v_fontconfig=2.17.1
 
 
 ## Dependency tree
@@ -27,7 +29,9 @@ dep_freetype2=()
 dep_fribidi=()
 dep_harfbuzz=()
 dep_unibreak=()
-dep_libass=(freetype2 fribidi harfbuzz unibreak)
+dep_libxml2=()
+dep_fontconfig=(libxml2)
+dep_libass=(freetype2 fribidi harfbuzz unibreak fontconfig)
 dep_lua=()
 dep_libplacebo=()
 dep_mpv=(ffmpeg libass lua libplacebo)
@@ -40,4 +44,4 @@ dep_mpv_android=(mpv)
 v_ci_ffmpeg=n8.0
 
 # filename used to uniquely identify a build prefix
-ci_tarball="prefix-ndk-${v_ndk}-lua-${v_lua}-unibreak-${v_unibreak}-harfbuzz-${v_harfbuzz}-fribidi-${v_fribidi}-freetype-${v_freetype}-mbedtls-${v_mbedtls}-ffmpeg-${v_ci_ffmpeg}.tgz"
+ci_tarball="prefix-ndk-${v_ndk}-lua-${v_lua}-unibreak-${v_unibreak}-harfbuzz-${v_harfbuzz}-fribidi-${v_fribidi}-freetype-${v_freetype}-libxml2-${v_libxml2}-fontconfig-${v_fontconfig}-mbedtls-${v_mbedtls}-ffmpeg-${v_ci_ffmpeg}.tgz"
--- a/buildscripts/include/download-deps.sh
+++ b/buildscripts/include/download-deps.sh
@@ -47,6 +47,20 @@ if [ ! -d unibreak ]; then
 		tar -xz -C unibreak --strip-components=1
 fi
 
+# libxml2
+if [ ! -d libxml2 ]; then
+	mkdir libxml2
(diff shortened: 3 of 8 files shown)


moi15moi:
The build fail because the [CI](https://github.com/moi15moi/mpv-android/blob/2a8f609adc8a9f3a8e540707c6b4cfe88fea0825/.github/workflows/build.yml#L29) doesn't seems to use ``buildscripts/include/download-sdk.sh`` to install the dependencies.
fontconfig require gperf to be installed, so I guess I will also need to edit the CI.
I'm not sure why the CI build is pretty different to the "normal" build we do locally.

Edit: Is there any reason why the CI doesn't produce a arm64 build? I believe most people have arm64 device, so it would be much more conveniant to test CI build.

moi15moi:
If you want, I can replace expat with libxml, so we could also fix https://github.com/mpv-android/mpv-android/issues/979

sfan5:
Not a fan of libxml, due to its history with security issues and it also briefly went unmaintained. Though if we have to include an XML library anyway then it would indeed be advantageous to also get DASH support working.
Then I take a look at [expat](https://github.com/libexpat/libexpat) and:
> Caution
> Expat is understaffed and without funding.

...

It also feels a bit overengineered to include fontconfig + an XML library just so it can do the equivalent of `ls /system/fonts` and not read any config file ever (because we don't ship one). :shrug: 

> Edit: Is there any reason why the CI doesn't produce a arm64 build? I believe most people have arm64 device, so it would be much more conveniant to test CI build.

No strong reason. The scripts used to require at least armv7l to be built before creating an APK, and it was also the most compatible arch.

moi15moi:
> so it can do the equivalent of `ls /system/fonts`

The problem to passing this directory to libass is that it will take too much memory. See: https://github.com/libass/libass/issues/548

sfan5:
Yeah it's unfortunate that libass doesn't support lazy-loading fonts from a big folder. There was also https://github.com/libass/libass/pull/312 but that went nowhere.

Anyway I've made up my mind. Can you change this PR to compile libxml2?
Then I'll do the ffmpeg integration to make DASH work later. And also get rid of `subfont`.

moi15moi:
Do we really want to remove `subfont`?
The default font will now not be the same on all android device and maybe people want to keep the original behaviour.

sfan5:
> Do we really want to remove `subfont`? The default font will now not be the same on all android device and maybe people want to keep the original behaviour.

Since Android consistently ships with Roboto I would expect fontconfig to always pick that by default, or will it not?

astiob:
Current libass will pick whichever font happens to be the first to support the requested glyph in the full list of fonts retrieved from Fontconfig.

With the Fontconfig-related changes included in https://github.com/libass/libass/pull/607, libass would instead ask Fontconfig to choose a font. I don’t know Fontconfig’s logic, but FWIW I see no reason for it to have a particular preference for Roboto above anything else. I would assume the result would at least be identical on all devices sharing the same set of installed fonts, though.

sfan5:
Sounds like mpv-android has to ship at least a basic fonts.conf so that font selection will behave broadly in-line with Android platform default fonts, then (the actual rules are [more complex](https://github.com/betacraft/Android-5.1-x86-rootfs/blob/master/system/etc/fonts.xml)).

moi15moi:
Creating a font.conf file can be pretty complicated.
From https://stackoverflow.com/a/39536066/15835974:

> A device manufacturer or custom ROM may change the default system fonts.

This means Roboto can be not present.
I believe it isnt the end of the world to not give a font.conf file so "serif", "sans-serif", "cursive", etc (see more generic family [here](https://github.com/libass/libass/issues/608#issuecomment-2664265517)) are mapped properly.

sfan5:
But then the "default font inconsistent" situation you warned of would happen.
It's also not unthinkable that fontconfig would pick an Asian font that isn't well suited for Latin text by default, if it has no guidance.

> Creating a font.conf file can be pretty complicated.

I'm aware. Maybe querying the default font (or `serif`, `sans-serif`, `monospace` each) from Java and dynanmically generating a very basic fonts.conf is realistic?

sfan5:
Small problem: initializing fontconfig (which happens at startup and sometimes during) takes a noticeable amount of time and also leads to frame drops
```
03-17 23:25:53.397 26758 26958 V mpv     : [sub/ass:v] Setting up fonts...
03-17 23:25:53.400 26758 26958 V mpv     : [sub/ass:info] No usable fontconfig configuration file found, using fallback.
03-17 23:25:53.864 26758 26958 V mpv     : [sub/ass:v] Using font provider fontconfig
^ 460ms

03-17 23:27:56.792 26758 26958 V mpv     : [osd/libass:v] Setting up fonts...
03-17 23:27:56.796 26758 26958 V mpv     : [osd/libass:info] No usable fontconfig configuration file found, using fallback.
03-17 23:27:57.287 26758 26958 V mpv     : [osd/libass:v] Using font provider fontconfig
^ 490ms
```

I expect this can be fixed by making sure fontconfig can write its cache correctly.

sfan5:
Merged as 7ece8b49714745bf9d10b270db96443400685749, thanks.

sfan5:
> I'm aware. Maybe querying the default font (or `serif`, `sans-serif`, `monospace` each) from Java and dynanmically generating a very basic fonts.conf is realistic?

So, this is possible using `AFontMatcher_match` and freetype (see 7930b4d354e2dd690a9d96469570018968138712), but pretty annoying so I'll not do it.

moi15moi:
It is indeed possible, but that require the api 29

sfan5:
Android 10 is 7 years old, so that's not the concern.
Anyway I went with hardcoding the standard font names to ensure the selection behaves somewhat predictably.

PR #1226, opened 2026-03-21
Implement playlist search / filtering

### LINKED ISSUE #368: search box when displaying playlist
hi, I think it would be nice to have a search bar for specially long item playlist like m3u. It's a pain to search all the list to find something that you have already in mind. a searchable "fuzzy find" like.

9mdv:
Fixes #368
This is my second attemp, the first one was using fuzzy search based on the main mpv project which works fine but quite slow and had UI issues in landscape mode. This is Gemini-assisted so feel free to point out any issue. Thanks.

<details>
  <summary>Details</summary>

# Playlist Search and Filtering Implementation

This document describes the reimplementation of the playlist search and filtering system in `mpv-android`, focusing on precision, performance, and UI consistency.

## Overview

The original filtering logic used a character-by-character fuzzy match which often produced imprecise results. The new implementation shifts to a word-based "contains all" approach, optimized with modern Android UI patterns.

### 1. Enhanced Filtering Logic (Precision)
- **Word-Based Matching**: The search query is now split into individual parts (tokens) based on whitespace.
- **"Contains All" Requirement**: For an item to match, *every* part of the search query must be found in either the `title` or the `filename` (case-insensitive).
- **Improved UX**: This behavior is more predictable for users than loose fuzzy matching, as it allows for filtering by multiple criteria (e.g., "05 live" to find track 5 from a live album).

### 2. Performance Optimizations
- **DiffUtil Integration**: Replaced the expensive `notifyDataSetChanged()` with `DiffUtil.calculateDiff`.
    - **Efficiency**: Only items that actually change are updated in the `RecyclerView`.
    - **Visuals**: Provides smooth, native animations when items are added/removed during filtering.
- **Search Debouncing**: Added a 200ms debounce using a `Handler` and `Runnable`.
    - **CPU Savings**: Prevents the filter logic from running on every single keystroke if the user is typing quickly.
    - **Stability**: Automatically cancels pending filter operations when a manual `refresh()` is triggered.

### 3. UI/UX Improvements (Landscape Mode)
- **Inline Search**: Applied `flagNoExtractUi` to the `SearchView` via both XML and Kotlin.
- **Behavior**: This prevents Android from showing a full-screen "extract" text editor in landscape mode. The user can now see the playlist update in real-time beneath the keyboard while typing in landscape orientation.

### 4. Code Robustness
- **Null Safety**: Implemented proper Kotlin null-safety checks for `Handler` callbacks to prevent crashes during rapid UI interactions.
- **State Management**: Ensured the `selectedIndex` (currently playing item) is correctly tracked and scrolled into view even after filtering.

## Implementation Files
- `app/src/main/java/is/xyz/mpv/PlaylistDialog.kt`: Core logic, `DiffUtil` callback, and debouncing.
- `app/src/main/res/layout/dialog_playlist.xml`: UI definition and IME options.
</details>


app/src/main/java/is/xyz/mpv/PlaylistDialog.kt
```diff
@@ -45,15 +54,60 @@ internal class PlaylistDialog(private val player: MPVView) {
...
+            override fun onQueryTextChange(newText: String?): Boolean {
+                filterRunnable?.let { handler.removeCallbacks(it) }
+                filterRunnable = Runnable { filter(newText) }
+                handler.postDelayed(filterRunnable!!, 200)
+                return true
+            }
+        })
+
+        // don't go full-screen in landscape
+        binding.searchView.imeOptions = EditorInfo.IME_FLAG_NO_EXTRACT_UI
```
sfan5, line 69: this is already set in the layout xml, isn't it?

app/src/main/res/values-ca/strings.xml
```diff
@@ -62,6 +62,7 @@
     <string name="file_picker_old">Selector de fitxers\n(heretat)</string>
     <string name="action_settings">Paràmetres</string>
     <string name="action_playlist">Llista de reproducció</string>
+    <string name="search_playlist">Seleccioneu una entrada de la llista</string>
```
sfan5, line 65: where do these translations come from? I assume you don't speak all of these languages

app/src/main/java/is/xyz/mpv/PlaylistDialog.kt
```diff
@@ -45,15 +54,60 @@ internal class PlaylistDialog(private val player: MPVView) {
...
+
         Utils.handleInsetsAsPadding(binding.root)
         return binding.root
     }
 
+    private fun filter(query: String?) {
+        val newPlaylist = if (query.isNullOrBlank()) {
+            fullPlaylist
+        } else {
+            val queryParts = query.lowercase().split(Regex("\\s+")).filter { it.isNotEmpty() }
```
sfan5, line 82: could short-circuit `queryParts.isEmpty()` here

Issue #1019, opened 2024-10-18
If a file fails to play it should show a popup


sfan5:
...with log messages from mpv.

use a `LogObserver` and in-memory buffer of errors
AlertDialog before closing

only do this for the first/last file to not obstruct playback too much?

sfan5:
alternative idea: show a ❗ somewhere at the top and you can then view the log errors in the top menu

PR #1233, opened 2026-04-10
Add export logs button in dev settings

moi15moi:
There are many people that open issue without the logs.
It happen because it isn't "easy" to get the logs. I believe that with this change, there will be less bug report without logs.

Note that this PR was assisted with AI.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -245,6 +245,16 @@
                 <action android:name="is.xyz.mpv.ACTION_NEXT" />
             </intent-filter>
         </receiver>
+
+        <provider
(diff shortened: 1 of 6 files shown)


sfan5:
I basically planned this anyway (#1019), but one reason I haven't done it specifically for issue logs is that the mpv logs might not include enough information ... however I see you're just calling out to `logcat` in the app.
That works? Isn't that basically a hole in Google's sandbox? :thinking:

moi15moi:
I mean, it does seems to work well.


> Isn't that basically a hole in Google's sandbox?

I guess? I havent tried, but maybe it isnt possible to call logcat to get logs from other app

PR #1240, opened 2026-04-25
MPVActivity: Simplify system bars and cutout inset handling

obscenelyvague:
Combine into single query instead of making this convoluted.


sfan5:
Will have to re-test, because this getting this right was very hard.
ref https://github.com/mpv-android/mpv-android/commit/2b16a044b996bf204d970b449f4cef33d2b0afdf

obscenelyvague:
Worth mentioning that this has been tested on a Pixel and Xiaomi device both running Android 15 with a display cutout and gesture navigation and a Samsung device running Android 6.0.1 with no display cutout and physical button navigation.

sfan5 (approved):

PR #1244, opened 2026-04-26
Misc fixes

obscenelyvague:


app/src/main/java/is/xyz/mpv/MPVActivity.kt
```diff
@@ -1309,7 +1309,7 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
     }
 
     private fun goIntoPiP() {
-        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.N)
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.O)
```
sfan5, line 1312: `enterPictureInPictureMode` was added in API level 24, which is N

app/src/main/java/is/xyz/mpv/TouchGestures.kt
```diff
@@ -71,7 +71,6 @@ internal class TouchGestures(private val observer: TouchGesturesObserver) {
     }
 
     fun setMetrics(width: Float, height: Float) {
-        assertFloat(width, height)
```
sfan5, line 74: the point of an assert is to catch unforeseen circumstances, so I want to keep it here

app/src/main/java/is/xyz/mpv/MPVActivity.kt
obscenelyvague: There is no PiP mode prior to API 26 though so not sure what that does in practice. Feel free to close this PR.

sfan5: :shrug: 
I merged the other commit already in case you didn't see, thanks.

PR #1249, opened 2026-04-29
Update Russian translation

solokot:


dairfod:
It's better to leave the "Дебандинг" as is.

solokot:
@dairfod,
In your opinion, why is the anglicism "Дебандинг" better if it is only understood by those who understand "Debanding"?

dairfod:
@solokot this anglicism is quite recognizable and has taken root. And it's more understandable to most people. Your version has a right to exist as well. I just want to avoid unnecessary confusion.

sfan5:
It would be nice if you could find a consensus, because I can't judge the translations.

dairfod:
Perhaps someone else could share their opinion. But this is a fairly well-established term. As a compromise, we could use something like:

Дебандинг (устранение полос)

But this clutters the interface.

solokot:
Since two different points of view have collided, let the anglicism "Дебандинг" remain in "pref_video_debanding_title" and let the explanation of the term from this pull request appear in "pref_video_debanding_summary".
We are unlikely to hear someone's third opinion here.

@sfan5, currently, I only have access to the web version of GitHub, so I cannot edit the pull. Will you be able to accept this pull, keeping the previous value of "pref_video_debanding_title", or should I close this one and create a new pull?

app/src/main/res/values-ru/strings.xml
```diff
@@ -124,8 +128,8 @@
 
     <string name="pref_video_downscale_title">Фильтр уменьшения масштаба</string>
 
-    <string name="pref_video_debanding_title">Дебандинг</string>
-    <string name="pref_video_debanding_summary">Выбор режима дебандинга видео.</string>
+    <string name="pref_video_debanding_title">Удаление полос</string>
```
sfan5:
```suggestion
    <string name="pref_video_debanding_title">Дебандинг</string>
```

PR #1250, opened 2026-04-30
app/build: Remove deprecated applicationVariants

obscenelyvague:
Tested on Gradle 9.5.0.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,3 @@
-# FIXME: the 'applicationVariants' stuff has changed
-android.newDsl=false
 android.useAndroidX=true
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,4 +1,3 @@
-# FIXME: the 'applicationVariants' stuff has changed
 android.newDsl=false
```
sfan5: if everything worked fine then this can be removed
```suggestion
```

obscenelyvague:
Rebased with flag removed. Thanks.

sfan5 (approved):
works, thanks

PR #1251, opened 2026-04-30
BackgroundPlaybackService: push thumbnail update on track changes

obscenelyvague:


sfan5:
To me it seems like `MPV_EVENT_VIDEO_RECONFIG` might happen more often than just track changes, is it a good idea to do it like this?

obscenelyvague:
When backgrounded? I'm open to other approaches of doing this if you have ideas.

sfan5:
Yes when backgrounded.
One idea: `MPV_EVENT_START_FILE` -> start a timer for 150ms (debounce) -> update thumbnail

Issue #697, opened 2023-04-26
Screen Brightness


shockergit:
How to make mpv save the last used screen Brightness as default for next playback

Issue #282, opened 2020-06-23
remember brightness, volume and playback speed


532910:
Please add controls to save brightness, volume level and playback speed

sfan5:
duplicate of #167, #168

532910:
Oh, I'm sorry!

Issue #264, opened 2020-05-29
[Feature Request] Ability to set a default device brightness when playing files (basically videos)


crisalis2:
I'm not talking about internal video brightness (mpv setting), but the device brightness you can change when playing a file with mpv-android by scrolling up and down in the left part of the screen.

crisalis2:
Sorry, I think this is a duplicate of #69

Issue #167, opened 2019-02-03
save brightness level


c-e-p-x-u-o:
Please add the control to save brightness level on exit and restore in on next time. Like VLC does.

blackCauldron7:
Are there plans to implement this?

07416:
`gamma` and `brightness` in the config file obviously don't affect screen brightness. Is there any [preference](https://mpv.io/manual/master/) which does?

PR #1255, opened 2026-05-03
MPVActivity: Remember screen brightness level set by user

obscenelyvague:
Some form of this feature has been requested many times over the years. Should match default behaviour of other players.

Adresses: #69, #167, #264, #282 and #697

### DIFF
--- a/app/src/main/java/is/xyz/mpv/MPVActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/MPVActivity.kt
@@ -178,6 +178,9 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
     private var showMediaTitle = false
     private var useTimeRemaining = false
 
+    private var rememberBrightness = false
+    private var lastScreenBrightness = -1
+
     private var ignoreAudioFocus = false
     private var playlistExitWarning = true
     private var newIntentReplace = false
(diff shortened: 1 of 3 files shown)


sfan5:
Hmm, I'm thinking this should have a setting. Personally I would disable this.
Is it common for other player apps to remember the brightness by default?

app/src/main/java/is/xyz/mpv/MPVActivity.kt
```diff
@@ -2083,7 +2096,10 @@ class MPVActivity : AppCompatActivity(), MPVLib.EventObserver, TouchGesturesObse
                 lp.screenBrightness = newBright
                 window.attributes = lp
 
-                gestureTextView.text = getString(R.string.ui_brightness, (newBright * 100).roundToInt())
+                val newBrightPercent = (newBright * 100).roundToInt()
+                gestureTextView.text = getString(R.string.ui_brightness, newBrightPercent)
+
+                lastScreenBrightness = newBrightPercent
```
sfan5: (replacing everything starting with `val lp`)
```kotlin
                val newBrightPercent = ((initialBright + diff) * 100).coerceIn(0f, 1f).roundToInt()
                lastScreenBrightness = newBrightPercent
                updateScreenBrightness()

                gestureTextView.text = getString(R.string.ui_brightness, newBrightPercent)
```

obscenelyvague:
>I'm thinking this should have a setting. Personally I would disable this.
Is it common for other player apps to remember the brightness by default?

From what I remember MX Player did this by default. Added user toggle. Untested.

obscenelyvague:
Should be `((initialBright + diff).coerceIn(0f, 1f) * 100)` not `((initialBright + diff) * 100).coerceIn(0f, 1f)`. Fixed, and now tested.

sfan5 (approved):

sfan5:
merged as cafc790ac7575d71d1fcbe3da3679cb7ebb3d07b, thanks.

PR #1260, opened 2026-05-10
Misc fixes II

obscenelyvague:


app/src/main/java/is/xyz/mpv/Utils.kt
```diff
@@ -509,10 +509,9 @@ internal object Utils {
         var originalPadding: Padding? = null
         ViewCompat.setOnApplyWindowInsetsListener(view) { _, insets ->
             val i = insets.getInsets(WindowInsetsCompat.Type.systemBars())
-            // yes, really
```
sfan5, line 512: the comment points toward just remembering the first padding values we see, which is a huge hack. feel free to reword but it should stay.

obscenelyvague: Kept the comment.

PR #1264, opened 2026-05-17
buildscripts: add curl

sfan5:


syphyr:
You can also disable ipfs and ldap support (--disable-{ipfs,ldap}).

```
--- a/buildscripts/scripts/curl.sh
+++ b/buildscripts/scripts/curl.sh
@@ -20,7 +20,8 @@ cd $build
        --host=$ndk_triple --with-mbedtls="$prefix_dir" --without-libpsl \
        --disable-shared --enable-static --disable-debug \
        --disable-{manual,docs,ares,unix-sockets,tls-srp,doh} \
-       --disable-{rtsp,dict,telnet,tftp,pop3,imap,smb,smtp,gopher,mqtt,ntlm}
+       --disable-{rtsp,dict,telnet,tftp,pop3,imap,smb,smtp,gopher,mqtt,ntlm} \
+       --disable-{ipfs,ldap}
 
 make -j$cores
 make DESTDIR="$prefix_dir" install
```

sfan5:
ldap won't be included anyway since the libraries are not present and I believe ipfs is just some trivial wrapper code (so not worth disabling).

buildscripts/scripts/curl.sh
```diff
@@ -0,0 +1,29 @@
...
+	--host=$ndk_triple --with-mbedtls="$prefix_dir" --without-libpsl \
+	--disable-shared --enable-static --disable-debug \
+	--disable-{manual,docs,ares,unix-sockets,tls-srp,doh} \
+	--disable-{rtsp,dict,telnet,tftp,pop3,imap,smb,smtp,gopher,mqtt,ntlm}
+
+make -j$cores
+make DESTDIR="$prefix_dir" install
+
+# curl adds a dependency which doesn't actually exist
+sed -re '/^Requires/ s|\bzlib\b| |g' -i "$prefix_dir/lib/pkgconfig/libcurl.pc"
```
moi15moi: Maybe you wouldnt get zlib if you would use ``--without-zlib`` for the configure ?
It seems to be caused by this: https://github.com/curl/curl/blob/014be82a66f62972a68c4c9ae5300edc84e9b0df/configure.ac#L1558-L1560

sfan5: I do want zlib, obviously. This was already reported and fixed in curl so the workaround will not be needed in the future.

syphyr:
The version of curl can also be bumped

PR #1285, opened 2026-07-28
Downloadsdkfixups

obscenelyvague:


obscenelyvague:
Werks on my machine ¯⁠\⁠_⁠(⁠ツ⁠)⁠_⁠/⁠¯
CI parts need rethinking

obscenelyvague:
Currently AGP doesn't find NDK for `Task :app:stripDefaultReleaseDebugSymbols` (only in CI)
This needs to be resolved.

buildscripts/include/ci.sh
```diff
@@ -49,12 +49,6 @@ if [ "$1" = "export" ]; then
 	exit 0
 elif [ "$1" = "install" ]; then
 	# install deps
-	if [[ -n "$ANDROID_HOME" && -d "$ANDROID_HOME" ]]; then
-		msg "Linking existing SDK"
-		mkdir -p sdk
-		ln -sv "$ANDROID_HOME" sdk/android-sdk-linux
-	fi
```
sfan5, line 56: I would like to continue re-using the SDK components from the runner.
why these changes at all?

obscenelyvague: This change was made because only the latest version of commandlinetools contain the new Android CLI binary. 

However, the culmination of all the changes are overall are a regression for CI workflow since it now ends up exhausting more bandwidth per run. sdkmanager will probably keep working fine for a while so I will close this PR as a whole.

PR #1292, opened 2026-08-09
jni: clear global reference to app context

obscenelyvague:


sfan5:
ffmpeg holds onto this reference indefinitely and there's no way to unset it, so I don't think we can do this actually:
https://github.com/FFmpeg/FFmpeg/blob/2a20737f662fd3f3f5999e873b9e7c90b5efc375/libavcodec/jni.c#L84-L107

obscenelyvague:
Can we inadvertently avoid the leak by passing app context instead of activity context here?
https://github.com/mpv-android/mpv-android/blob/b44d3addc2a6bfb140a862308955a836e4a922d2/app/src/main/java/is/xyz/mpv/BaseMPVView.kt#L18

PR #1294, opened 2026-08-11
jni: clear references at method cache init

obscenelyvague:


sfan5:
The [JNI docs](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/design.html) say:

> In most cases, the programmer should rely on the VM to free all local references after the native method returns. However, there are times when the programmer should explicitly free a local reference. Consider, for example, the following situations:
>
>    * A native method accesses a large Java object, thereby creating a local reference to the Java object. The native method then performs additional computation before returning to the caller. The local reference to the large Java object will prevent the object from being garbage collected, even if the object is no longer used in the remainder of the computation.
>    * A native method creates a large number of local references, although not all of them are used at the same time. Since the VM needs a certain amount of space to keep track of a local reference, creating too many local references may cause the system to run out of memory. For example, a native method loops through a large array of objects, retrieves the elements as local references, and operates on one element at each iteration. After each iteration, the programmer no longer needs the local reference to the array element.

So I'm kinda wondering if this isn't pointless actually?

obscenelyvague:
https://developer.android.com/ndk/guides/jni-tips#local-and-global-references
Seems to contradict that? 
>Programmers are required to "not excessively allocate" local references. In practical terms this means that if you're creating large numbers of local references, perhaps while running through an array of objects, you should free them manually with DeleteLocalRef instead of letting JNI do it for you. The implementation is only required to reserve slots for 16 local references, so if you need more than that you should either delete as you go or use EnsureLocalCapacity/PushLocalFrame to reserve more.

If methods are added in the future and surpass 16 references, would it then not be required?

sfan5:
I don't think that's a contradiction. The Android guide just explains the limitations in more detail.

So yes, if we ever add more classes I would just go and call `EnsureLocalCapacity`.

PR #1302, opened 2026-08-15
PreferenceActivity: show section name as title instead of subtitle

### LINKED ISSUE #1152: Top app bar displays overlapping titles
### Important information

Android version:13

mpv-android version:mpv-android 2025-08-25-release / 8043 (release)

Which version of mpv-android introduced the problem (if known):..

### Description

When opening any settings section (e.g., Video, General, etc.), the top app bar doesn’t display correctly. It currently shows something like:

Image

with the second line (“Video”) partially cut off. Ideally, only the name of the opened section (Video) should appear in the top bar similar to how it works correctly in the About MPV section.

Image

This issue has been present for a long time. It’s a minor (low-priority) UI bug, but fixing it would make the settings interface more consistent and polished.

4eh5xitv6787h645ebv:
The settings screen used ActionBar title + subtitle, and the resulting two-line bar gets clipped or overlaps with large system fonts on some devices (#1152 has screenshots from MIUI and Moto).

This shows only the current section name, kept as the back stack entry name so the title is always derived from the actual back stack state: it survives rotation and process death (FragmentManager persists entry names), and the listener is registered in onStart/onStop with a resync, which also fixes a pre-existing bug where the subtitle went stale after stop/restart (the old listener was removed in onSaveInstanceState and never re-added).

Tested on a Pixel 3 (Android 15): root shows "Settings", sections show their name, title stays correct across home-and-return, back navigation, and rotation.

fixes #1152

### DIFF
--- a/app/src/main/java/is/xyz/mpv/preferences/PreferenceActivity.kt
+++ b/app/src/main/java/is/xyz/mpv/preferences/PreferenceActivity.kt
@@ -27,7 +27,6 @@ class PreferenceActivity : AppCompatActivity(),
 
         preferences = PreferenceManager.getDefaultSharedPreferences(this)
         preferences.registerOnSharedPreferenceChangeListener(this)
-        supportFragmentManager.addOnBackStackChangedListener(this)
         if (preferences.getBoolean("material_you_theming", false))
             DynamicColors.applyToActivityIfAvailable(this)
         enableEdgeToEdge()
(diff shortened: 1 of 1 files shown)


4eh5xitv6787h645ebv:
Sorry, this wasn't meant to be submitted yet — was doing some testing on my end. Closing, apologies for the noise.

PR #1313, opened 2026-08-27
filepicker: add natural sort order toggle (fixes #1295)

### LINKED ISSUE #1295: File picker: option for natural sort order (persisted toggle)
Right now the file picker sorts entries lexicographically, so files like `10.m4a` end up listed before `2.m4a`. Would be great to have a natural sort option (where numeric substrings are compared by value, not character-by-character) as an alternative.

Suggestion for the UX: a toggle next to the existing media-file filter icon (top right), persisted the same way the filter state already is (`MainActivity_filter_state` in SharedPreferences) — so users could switch between "alphabetical" and "natural" sort and have it stick across sessions.

From a quick look at the code, this seems fairly contained: `AbstractFilePickerFragment` (from NoNonsense-FilePicker) exposes `compareFiles(File lhs, File rhs)` as an overridable method, and mpv-android's `MPVFilePickerFragment` already extends it — so the comparator swap could live entirely there without touching native code.

Happy to help test if someone picks this up. Thanks for maintaining this app, it's still the most reliable local/network player I've found on Android!

Maqbool61:
## Problem

The file picker sorts entries lexicographically, which means files like
`10.mp4` appear before `2.mp4`. This is confusing for media libraries
with numbered tracks or episodes.

Fixes #1295.

## Solution

Add a persisted natural sort toggle to the file picker toolbar, next to
the existing media-file filter icon.

Natural sort compares numeric substrings by value rather than
character-by-character, so `2.mp4` correctly appears before `10.mp4`.
Leading zeros are handled (`01` == `1`). The preference persists across
sessions in SharedPreferences under the key `MainActivity_sort_natural`,
mirroring how the existing filter state is stored.

## Changes

- `MPVFilePickerFragment` — override `compareFiles()` with a
  `naturalCompare()` implementation; add `naturalSort` flag
- `MPVDocumentPickerFragment` — same pattern for the document picker path
- `DocumentPickerFragment` — replace hardcoded `Collections.sort(files)`
  with a `compareDocuments()` hook that subclasses can override
- `FilePickerActivity` — handle the new menu item, persist/restore sort
  state alongside the existing filter state
- `menu_filepicker.xml` — add `action_sort_order` item
- `ic_sort_24dp.xml` — new sort icon (Material style, matches filter icon)
- `strings.xml` — add `action_toggle_sort`, `notice_sort_natural`,
  `notice_sort_alpha`

## Screenshots
Screenshot_2026-08-27_17-37-12
Screenshot_2026-08-27_17-37-19

| Alphabetical (default) | Natural sort |
|---|---|
| `10.mp4` → `2.mp4` | `2.mp4` → `10.mp4` |

*(see attached screenshots)*

## Testing

1. Open file picker in a folder with numbered files (e.g. `1.mp4`,
   `2.mp4`, `10.mp4`, `11.mp4`)
2. Default order is alphabetical — `10.mp4` sorts before `2.mp4`
3. Tap the sort icon in the toolbar → toast: "Natural sort order"
4. List immediately re-sorts: `2.mp4` now before `10.mp4`
5. Force-stop and relaunch — sort preference is preserved
6. Tap again → toast: "Alphabetical sort order" — reverts correctly


app/src/main/java/is/xyz/mpv/FilePickerActivity.kt
```diff
@@ -91,8 +86,6 @@ class FilePickerActivity : AppCompatActivity(), AbstractFilePickerFragment.OnFil
     private fun doUiTweaks() {
         WindowCompat.setDecorFitsSystemWindows(window, false)
 
-        // Part 2 of the workaround: apply the insets to the recycler so it can
-        // take them into account.
```
sfan5, line 95: presumably there is no reason at all why you removed these and other comments?

Maqbool61: Sorry about that, those comments were accidentally stripped when I resolved 
the merge conflict. Fixing now.

app/src/main/java/is/xyz/mpv/FilePickerActivity.kt
```diff
@@ -111,6 +111,31 @@ class FilePickerActivity : AppCompatActivity(), AbstractFilePickerFragment.OnFil
...
+    private fun applySortState(natural: Boolean) {
+        fragment?.let {
+            it.naturalSort = natural
+            it.goToDir(it.getCurrentDir())
+        }
+        fragment2?.let {
+            it.naturalSort = natural
+            it.goToDir(it.getCurrentDir())
+        }
+    }
```
sfan5, line 137: doesn't hurt of course, but is there any reason why someone would want to disable this?

app/src/main/java/is/xyz/mpv/MPVDocumentPickerFragment.java
```diff
@@ -9,10 +9,61 @@
...
+            } else {
+                int cmp = Character.toLowerCase(ca) - Character.toLowerCase(cb);
+                if (cmp != 0) return cmp;
+                i++;
+                j++;
+            }
+        }
+        // compare remaining (unconsumed) character counts, not total lengths
+        return (a.length() - i) - (b.length() - j);
+    }
```
sfan5: this should be deduplicated.
doesn't have to be super nice and clean, e.g `MPVFilePickerFragment` could just make its method public

Maqbool61:
Done, naturalCompare is now public static in MPVFilePickerFragment and MPVDocumentPickerFragment delegates to it.

PR #1316, opened 2026-08-29
ci: Produce arm64 apk

### LINKED ISSUE #1287: ARM64 debug apk artifacts
Would you mind adding ARM64 debug APK artifacts to your future workflows? Currently, only an armv7 APK is included.

moi15moi:
Fix https://github.com/mpv-android/mpv-android/issues/1287

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,5 +43,10 @@ jobs:
 
       - uses: actions/upload-artifact@v7
         with:
-          name: mpv-android
+          name: mpv-android-armeabi-v7a
           path: app/build/outputs/apk/default/debug/app-default-armeabi-v7a-debug.apk
+
+      - uses: actions/upload-artifact@v7
+        with:
+          name: mpv-android-arm64-v8a
+          path: app/build/outputs/apk/default/debug/app-default-arm64-v8a-debug.apk
--- a/buildscripts/include/ci.sh
+++ b/buildscripts/include/ci.sh
@@ -5,6 +5,8 @@ cd "$( dirname "${BASH_SOURCE[0]}" )/.."
 
 . ./include/depinfo.sh
 
+v_ci_archs="armv7l arm64"
+
 msg() {
 	printf '==> %s\n' "$1"
 }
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -44,4 +44,6 @@ jobs:
       - uses: actions/upload-artifact@v7
         with:
           name: mpv-android
-          path: app/build/outputs/apk/default/debug/app-default-armeabi-v7a-debug.apk
+          path: |
+            app/build/outputs/apk/default/debug/app-default-armeabi-v7a-debug.apk
+            app/build/outputs/apk/default/debug/app-default-arm64-v8a-debug.apk
```
sfan5: should be separate artifacts

buildscripts/include/depinfo.sh
```diff
@@ -44,5 +44,8 @@ dep_mpv_android=(mpv)
 # pinned ffmpeg revision
 v_ci_ffmpeg=n9.0
 
+# architectures built by the CI (space-separated, see buildall.sh --arch)
+v_ci_archs="armv7l arm64"
+
 # filename used to uniquely identify a build prefix
-ci_tarball="prefix-n${v_ndk}-l${v_lua}-u${v_unibreak}-h${v_harfbuzz}-fr${v_fribidi}-ft${v_freetype}-x${v_libxml2}-fo${v_fontconfig}-m${v_mbedtls}-c${v_curl}-ff${v_ci_ffmpeg}.tgz"
+ci_tarball="prefix-n${v_ndk}-l${v_lua}-u${v_unibreak}-h${v_harfbuzz}-fr${v_fribidi}-ft${v_freetype}-x${v_libxml2}-fo${v_fontconfig}-m${v_mbedtls}-c${v_curl}-ff${v_ci_ffmpeg}-a${v_ci_archs// /-}.tgz"
```
sfan5: I know it's not "proper" but I think the tarball name is long enough, so just leave this unmodified and I'll delete the cache once I merge this

.github/workflows/build.yml
moi15moi: Ok, I named them `mpv-android-armeabi-v7a` + `mpv-android-arm64-v8a`.
If you want, i can add the parameter `archive` to false, so when we download the artifact, we directly download the apk instead of a zip file.

buildscripts/include/depinfo.sh
moi15moi: Ok, done.

.github/workflows/build.yml
sfan5: sure

sfan5:
Merged as 9714f15ce11e1364eea4330da390d357a22d2bb1, thanks.

.github/workflows/build.yml
moi15moi: Thanks for doing it.
Kinda sucks that archive false doesnt respect the name, but the bug have been reported: https://github.com/actions/upload-artifact/issues/785

PR #1326, opened 2026-09-09
external display support [closes #728]

### LINKED ISSUE #728: Proper HDMI Output
### Describe the wanted behaviour of the new feature and potential usecases for it here.

When connected to HDMI, output the video to the display connected for accurate screen resolution of desired video. This will allow for devices that do not have conventional screen resolutions to output video without black bars to modern TVs.

### Alternative behavior of the wanted feature
N/A

VLC does this, but their implementation for subs is questionable for anime which is why I would love to see this on MPV

simon-weber:
Lets video play on an external display via `android.app.Presentation`, so the phone screen can turn off without stopping playback. The phone shows the usual control ui, times out to just a black screen, and comes back when tapped.

Controlled by a new setting `General -> External display projection` on by default.

### Implementation

`ExternalDisplayManager` watches for a presentation-capable display and shows/dismisses a bare `SurfaceView` Presentation on it. Since libmpv only has one video output target, a new `MPVSurfaceCoordinator` arbitrates between the phone's surface and the Presentation's, doing a clean detach/attach on handoff. Projecting is treated like `background_play=always` (won't pause on backgrounding, starts the background service + a wake lock).

Code written by sonnet 5 high and tested manually. I read it over but don't have experience with android or this codebase.

### Testing

Tested on a OnePlus 15 with xreal xbx a01+ glasses in mirroring mode:
* plug in / unplug mid-playback
* cold start with glasses already connected
* settings toggle
* backgrounding via home button

Note that locking the phone blanks both displays.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -13,6 +13,7 @@
     <uses-permission android:name="android.permission.INTERNET" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
     <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PLAYBACK" />
+    <uses-permission android:name="android.permission.WAKE_LOCK" />
 
     <application
         android:allowBackup="true"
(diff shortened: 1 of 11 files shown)


pixel4696:
@sfan5 this also works nicely on other hdmi, really hope this could be merged soon!

